PR #14, opened 2021-02-23
use sa for kubernetes driver

simongdavies:
This change sets the service account for the Kubernetes Driver so that the installation image can connect to the Kubernetes cluster that it is running in


carolynvs:
Let's discuss the ARC scenario offline and then start with an issue describing how we want to support that. What's in this PR isn't the right long term fix because it assigns too many cluster roles to the invocation image.

PR #22, opened 2021-03-19
Implement Installation Service Account

### LINKED ISSUE #16: Add invocation image service account to AgentConfig
The bundle's invocation image may need to use in-cluster connections to the current cluster, in which case the service account for the invocation image needs to be set. The k8s driver supports passing that in, so we just need to make it configurable and pass it through.

AgentConfig can be specified directly on an installation, in the namespace of the installation or in the porter operator namespace.

simongdavies:
Adds implementation service account to agent config closes #16

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -16,7 +16,7 @@ type InstallationSpec struct {
 
 	// AgentConfig overrides the Porter Agent configuration defined at the namespace or system level.
 	// +optional
-	AgentConfig AgentConfigSpec `json:"agentConfig,omitEmpty"` // TODO: Make this a reference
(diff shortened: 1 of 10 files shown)


api/v1/installation_types.go
```diff
@@ -16,7 +16,7 @@ type InstallationSpec struct {
 
 	// AgentConfig overrides the Porter Agent configuration defined at the namespace or system level.
 	// +optional
-	AgentConfig AgentConfigSpec `json:"agentConfig,omitEmpty"` // TODO: Make this a reference
+	AgentConfig AgentConfigSpec `json:"agentConfig,omitempty"` // TODO: Make this a reference
```
carolynvs, line 19: 👍

carolynvs (approved):
LGTM, the code coverage isn't smart enough to see that the new code paths are tested. 🤷‍♀️

PR #31, opened 2021-05-17
Default to kubernetes plugin

simongdavies:
This change updates the instalation to use the kubernetes-plugin as the default instead of the azure plugin.


installer/porter.yaml
```diff
@@ -56,18 +61,23 @@ credentials:
     description: Kubeconfig file for cluster where the operator should be installed
     path: /root/.kube/config
   - name: config.toml
-    description: Porter configuration file found in ~/.porter/config.toml
+    required: false
```
carolynvs, line 64: You had it right for the installation service account parameter. In Porter if you want something to be optional, just set a default. In this context required doesn't do what you think: it is the list of required properties on the json schema for the parameter (used when the parameter is a json document).

So these should all be `default: ""` instead.

carolynvs (changes requested):
Everything is good, just need to tweak how the azure parameters are marked as optional.

carolynvs (approved):
SHIP. IT. :shipit:

PR #55, opened 2021-11-18
Opinionated Dockerfile && Golang Update

Church13:
- Opinionated changes to dockerfile -- specifically, using the `/app` folder as a workdir, and specifying the full path to the manager executable for the entrypoint directive
- Update golang build image from 1.15 tag to 1.17 to prevent building with a version of golang affected by CVE-2021-39293

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1.2
 
 # Build the manager binary
-FROM golang:1.15 as builder
+FROM golang:1.17 as builder
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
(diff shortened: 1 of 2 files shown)


carolynvs (changes requested):
I just tried this out locally by running `mage build deploy` and it's failing because the manager manifest specifies the command to run (it was the suggested template from the operator sdk).

Can you fix this line and then it should be good to go?

https://github.com/getporter/operator/blob/a4a019011017efa8c2e2f9958e7536d82f21c751/config/manager/manager.yaml#L32

carolynvs (approved):
Thanks for the fixes! 💯

Issue #58, opened 2022-01-14
Support open telemetry


carolynvs:
Porter has added initial support for open telemetry. The operator should pass through ambient opentel env vars to the porter agent so that we can collect trace data from it. Also the controller should be instrumented.

carolynvs:
We should generate a correlation id and pass it to the porter agent. See https://github.com/getporter/porter/issues/1873

carolynvs:
To clarify the scope here, this is passing through existing environment variables that the sysadmin may have set on the cluster but is not adding instrumentation to the operator code itself.

AGMETEOR:
Hey @schristoff can I get assigned this one.

PR #59, opened 2022-01-14
Implement uninstall

### LINKED ISSUE #27: Uninstall installation when CRD is deleted
When the CRD for an installation is deleted, uninstall the installation.

### LINKED ISSUE #28: Populate installation CRD status object
* last known status of the installation

### LINKED ISSUE #40: Handle deleting resources
All resources should have a finalizer registered, so that when it is deleted we have a chance to execute Porter quick to really remove the resource.

https://book.kubebuilder.io/reference/using-finalizers.html

For credential/parameter sets, we just need to call `porter credential|parameter delete NAME`. For installations, we need to call uninstall and wait for it to finish.

### LINKED ISSUE #46: Support uninstall without removing the CRD
In the Porter community meeting today, it came up that some people will want to use Kubernetes as the way to tell what is installed, and also _what used to be installed_. Right now the design is to uninstall when the CRD is deleted. To support this other way of managing resources, we could add a field to the installation indicating if it's installed or not.

**existing installation**
```yaml
kind: Installation
metadata:
  name: mywordpress
spec:
  name: wordpress
  bundle:
    repository: getporter/wordpress
    version: 0.4.0
status:
  installedOn: 2021-10-01... # Timestamp of when the bundle was installed
```

**installation that has been uninstalled**
```yaml
kind: Installation
metadata:
  name: mywordpress
spec:
  uninstalled: true # Add this to the spec to uninstall
  name: wordpress
  bundle:
    repository: getporter/wordpress
    version: 0.4.0
status:
  installedOn: 2021-10-01
  uninstalledOn: 2021-10-10... # Timestamp of when it was uninstalled
```

carolynvs:
Uninstall is now triggered when a CRD is deleted OR if spec.uninstalled=true. In either case we have the porter-agent call `porter installation apply`. This requires an unreleased build of porter with support for uninstalling a bundle by changing the document See https://github.com/getporter/porter/pull/1851. 

I have also updated our logging to use log.V(level) to make it easier to filter to the type of info. 5=trace, 4=debug, 3=system state, 2=app state, 1=info, 0=error. I borrowed this from another blog/k8s project but am having trouble finding it. I'm using constants so if we need to switch what the levels mean we can later. Similarly I have added trace statements so we can figure out what happened when things go wrong. Eventually we need to add instrumentation (see #58 ).

This PR has lots of general improvements to the operator that became relevant while implementing uninstall:
* Use a well-known annotation to trigger reconcile again: `porter.sh/retry`. We are now adding that value as a label to the agent job so that we can retry without accidentally picking up the previous job if it's still there.
* Use generation instead of resource version to detect changes. Resource version is updated anytime the entire installation is changed, while generation only updates when the spec changes or when the resource is deleted.
* Populate the installation status indicating if we have scheduled an agent, it's running, succeeded or failed. Emiting events will be in a separate PR. There isn't more detailed information about how the run went or what action porter took. For simplicity, and to avoid data sync issues with Porter's database, the installation status is about the porter agent job only.
* We reset the status when a new change is made to the installation (we aren't keeping run history for now).
* Split out the reconcile logic into smaller unit-tested functions. These tests check that we are creating resources properly and that reconcile only makes a single change then requeues.
* The integration test used to check a lot of details but since it's easier to check in unit tests, and the integration test is more for "does the whole thing work?" I've simplified the checks in the integration test.

🚨 This relies on https://github.com/getporter/porter/pull/1851 and https://github.com/getporter/porter/pull/1866. I will update our go reference to use that once it's released before merging.

Closes #27 
Closes #46 
Closes #28 
Closes #40


api/v1/installation_types.go
```diff
@@ -41,6 +41,9 @@ type InstallationSpec struct {
 	// Namespace (in Porter) where the installation is defined.
 	Namespace string `json:"namespace" yaml:"namespace"`
 
+	// Active specifies if the bundle should be installed.
+	Active bool `json:"active" yaml:"active"`
```
carolynvs, line 45: After this PR, I'll see if there's a better way that doesn't involve replicating the fields from the installation document used by porter. I think what needs to happen is splitting up Porter's document like the k8s manifests are:

Installation
* InstallationSpec (embedded so it's top level)
* InstallationStatus

Then we can embed a porter.InstallationSpec type in ours here and not have to duplicate everything.

api/v1/installation_types.go
```diff
@@ -100,11 +111,68 @@ func (in InstallationSpec) MarshalYAML() (interface{}, error) {
...
+
+	// PhaseFailed means that calling Porter failed.
+	PhaseFailed InstallationPhase = "Failed"
+)
+
+// These are valid conditions of an Installation.
+type InstallationConditionType string
+
+const (
+	// RunScheduled means that the Porter run has been scheduled.
```
vdice, line 163: Should the godoc comments be updated for these `Condition*` consts?

api/v1/installation_types.go
```diff
@@ -117,6 +185,21 @@ type Installation struct {
 	Status InstallationStatus `json:"status,omitempty"`
 }
 
+// BuidRetryLabel returns a value that is safe to use
```
vdice, line 188:
```suggestion
// GetRetryLabelValue returns a value that is safe to use
```

vdice (approved):

PR #73, opened 2022-02-16
Add AgentAction controller

### LINKED ISSUE #39: How to support invoke through the operator?
The new model for the operator is based around calling the porter apply commands that are being introduced in v1. This allow us to support the following options:

* install when a CRD is added
* upgrade when a CRD is changed
* uninstall when a CRD is deleted

Since the new model does not support imperative commands, the questions I have are:

* Do we need to? Is calling one off commands on an installation something that we need to support? Or like the Helm operator, should we just stick to the installation lifecycle, and otherwise people should continue to use the porter cli for one-offs?
* If we did support it, what would that look like? One thought would be to have a Run CRD that models executing any arbitrary porter command.

For our first iteration of the operator, @squillace has encouraged us to consider this out of scope but I wanted to get the discussion out there so others can provide feedback.

carolynvs:
I've split the installation controller in two, adding a new controller just to manage running the porter agent. The installation controller now only is responsible for creating an AgentAction resource. This triggers the AgentAction controller to make a porter agent job to execute the command from the AgentAction.

It's a bit of an extra layer of indirection but it will simplify additional controllers since they don't need to manage the porter agent job. It also enables users to request that the porter agent is run and can specify arbitrary commands, such as invoke.

Closes #39 

This requires that you have porter installed from a custom build of porter with support for [nonroot invocation images](https://github.com/getporter/porter/pull/1930). The build won't pass until that PR is merged and a new tag created so we can reference it.

TODO
- [x] Get off of the porter fork with "reusable document types" Just embed the fields for now and we can get fancy later.
- [x] Fix integration test and create follow-up #76


config/manager/kustomization.yaml
```diff
@@ -11,8 +11,3 @@ configMapGenerator:
 - files:
   - controller_manager_config.yaml
   name: manager-config
-
-images:
```
carolynvs, line 15: Setting the image of the controller now happens in the bundle itself so that it can use the image distributed with the bundle.

config/samples/porter-hello.yaml
```diff
@@ -9,5 +9,7 @@ spec:
   namespace: operator
   name: hello
   bundle:
-    repository: getporter/porter-hello
-    version: 0.1.1
+    repository: carolynvs/porter-hello-nonroot:v0.1.0
```
carolynvs, line 12: Switch to a bundle that both takes a parameter and generates outputs to pick up on file permission problems with running as nonroot

config/samples/porter-hello.yaml
```diff
@@ -9,5 +9,7 @@ spec:
   namespace: operator
   name: hello
   bundle:
-    repository: getporter/porter-hello
-    version: 0.1.1
+    repository: ghcr.io/getporter/test/porter-hello
```
carolynvs, line 12: This switches to a bundle that was built with the latest nonroot changes.

magefile.go
```diff
@@ -139,10 +146,11 @@ func getMixins() error {
 
 	mixins := []struct {
 		name    string
+		url     string
 		feed    string
 		version string
 	}{
-		{name: "helm3", feed: "https://mchorfa.github.io/porter-helm3/atom.xml", version: "v0.1.14"},
+		{name: "helm3", url: "https://github.com/carolynvs/porter-helm3/releases/download", version: "v0.1.15-8-g864f450"},
```
carolynvs, line 153: This is a build of helm3 that includes a patch for running in a nonroot bundle. https://github.com/MChorfa/porter-helm3/pull/42 I'll submit that PR when we've fully vetted the nonroot change.

magefile.go
```diff
@@ -174,44 +188,49 @@ func Publish() {
...
 
 // Generate k8s manifests for the operator.
 func buildManifests() {
 	mg.Deps(EnsureKustomize, EnsureControllerGen)
 
-	img := resolveControllerImage()
-	kustomize("edit", "set", "image", "manager="+img).In("config/manager").Run()
+	// Set the image reference in porter.yaml so that the manager image is packaged with the bundle
+	managerRef := resolveManagerImage()
+	mgx.Must(shx.Copy("installer/vanilla.porter.yaml", "installer/porter.yaml"))
```
carolynvs, line 203: I can't wait for this to be settable by a custom build arg! 😁

bdegeeter (commented):
I did a manual test (by accident) with the wrong version for a bundle in an installation resource.  I could not delete the resource via `kubectl delete -f`

Worked as expected when I fixed the version.

config/samples/_v1_agentaction.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: porter.sh/v1
+kind: AgentAction
+metadata:
+  name: agentaction-sample
+spec:
+  args: ["installation", "apply", "installation.yaml"]
+  files:
+    # base64 encoded file contents
+    installation.yaml: c2NoZW1hVmVyc2lvbjogMS4wLjAKbmFtZXNwYWNlOiBvcGVyYXRvcgpuYW1lOiBoZWxsbwpidW5kbGU6CiAgcmVwb3NpdG9yeTogZ2hjci5pby9nZXRwb3J0ZXIvdGVzdC9wb3J0ZXItaGVsbG8KICB2ZXJzaW9uOiAwLjIuMApwYXJhbWV0ZXJzOgogIG5hbWU6IGxsYW1hcyAK
```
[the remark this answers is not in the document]
carolynvs: You can always delete a resource by editing it and removing any finalizers on it.

I went with base64 because it allows people to add any files they need into the working directory of the agent, without us having known about it up-front. Do you have a suggestion for how you'd like to see it done differently while still being flexible?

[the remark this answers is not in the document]
carolynvs: I am concerned that it may be an indentation yaml bug generator because it isn't embedding a yaml document, it's a formatted yaml string embeded in a yaml file. Though either way it makes it much easier to introduce bugs in the document.

Since k8s users are already familiar with how to decode secrets, I am not too worried about the UX of doing exactly what secrets do for this. Let's go with following k8s existing conventions for now and if we need to revisit in later versions, I'm fine with that. I'd rather get this merged so we can try out a working operator and move forward with more controllers.

api/v1/const.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// LabelResourceGeneration is a label applied to the resources created by the
+	// Porter Operator, representing the generation of the owning resource. It is
+	// used to help the operator determine if a resource has
+	// already been created.
+	LabelResourceGeneration = Prefix + "resourceGeneration"
+
+	// LabelRetry is a label applied to the resources created by the
+	// Porter Operator, representing the retry attempt identifier. It is used to
+	// help the operator determine if a resource has
```
vdice, line 62: Looks like the comment is cut off here

controllers/installation_controller.go
```diff
@@ -112,55 +56,67 @@ func (r *InstallationReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	// Should we uninstall the bundle?
-	if shouldUninstall(inst) {
+	if r.shouldUninstall(inst) {
 		err = r.uninstallInstallation(ctx, log, inst)
 		log.V(Log4Debug).Info("Reconciliation complete: A porter agent has been dispatched to uninstall the installation.")
 		return ctrl.Result{}, err
 	} else if isDeleted(inst) {
 		// This is installation without a finalizer that was deleted
-		// We remove the finalizer after we successfully uninstall (or someone is manually cleaning things up)
+		// We remove the finalizer af
```
vdice, line 111: This just strikes me as an odd comment break point

vdice (approved):
Defer to @VinozzZ for final approval per outstanding feedback, otherwise LGTM

controllers/agentaction_controller.go
```diff
@@ -0,0 +1,704 @@
...
+			"namespace", config.Namespace,
+			"name", config.Name)
+	}
+
+	// Read agent configuration defined at the system level
+	systemCfg := &porterv1.AgentConfig{}
+	err := r.Get(ctx, types.NamespacedName{Name: "default", Namespace: operatorNamespace}, systemCfg)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return porterv1.AgentConfigSpec{}, errors.Wrap(err, "cannot retrieve system level porter agent configuration")
+	}
```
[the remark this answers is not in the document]
carolynvs: There is nothing for the controller to do. It's not an error case. I can add a comment about that.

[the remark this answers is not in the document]
carolynvs: Oh sorry I misunderstood which part of the code you were asking about. In this case, we are checking if there is a config defined, and using it if present. If it's not there, then that's not an error case. It just means we will use the defaults later on in that function.

PR #82, opened 2022-03-22
PorterConfigSpec ToPorterDocument doesn't honor optional fields

sgettys:


sgettys:
Failing UT to cover optional fields behavior for PorterConfigSpec

api/v1/porterconfig_types_test.go
```diff
@@ -40,33 +39,118 @@ func TestPorterConfigSpec_MergeConfig(t *testing.T) {
...
-		Secrets: []SecretsConfig{
-			{PluginConfig{
-				Name:         "keyvault",
-				PluginSubKey: "azure.keyvault",
-				Config:       runtime.RawExtension{Raw: []byte(`{"vault": "mysecrets"}`)},
-			}},
+		{
+			name: "Minimum fields set",
+			cfg: PorterConfigSpec{
+				DefaultStoragePlugin: pointer.StringPtr("mongodb"),
```
carolynvs, line 99: While this is a technically valid Porter configuration (i.e. it validates against our schema), know that it isn't a configuration that is capable of working currently. The mongodb plugin needs configuration in the operator to say which db to connect to.

api/v1/porterconfig_types_test.go
```diff
@@ -40,33 +39,118 @@ func TestPorterConfigSpec_MergeConfig(t *testing.T) {
...
+				DefaultSecretsPlugin: pointer.StringPtr("kubernetes.secrets"),
+			},
+			expDocument: []byte(`default-storage-plugin: mongodb
+default-secrets-plugin: kubernetes.secrets
+`),
+		},
+		{
+			name: "Storage Config minimum fields set",
+			cfg: PorterConfigSpec{
+				DefaultStoragePlugin: pointer.StringPtr("mongodb"),
```
carolynvs, line 109: You would never create a config like this because DefaultStorage takes precedence over DefaultStoragePlugin. For this test case, I think removing default storage plugin would yield the test case that you are still looking for (leaving the config section empty), no?

I realize we are just testing the schema here, but I think it helps to use examples that could work if used.

api/v1/porterconfig_types_test.go
```diff
@@ -40,33 +39,118 @@ func TestPorterConfigSpec_MergeConfig(t *testing.T) {
...
+storage:
+    name: in-cluster-mongodb
+		plugin: mongodb
+`),
+		},
+		{
+			name: "Secrets Config minimum fields set",
+			cfg: PorterConfigSpec{
+				DefaultStoragePlugin: pointer.StringPtr("mongodb"),
+				DefaultSecretsPlugin: pointer.StringPtr("kubernetes.secrets"),
```
carolynvs, line 131: Same for this test case, I recommend leaving off DefaultSecretsPlugin.

api/v1/porterconfig_types_test.go
sgettys: This is the question I had about "omitempty". For DefaultStoragePlugin "omitempty" is only set for json but not yaml or map structure so not providing this ends up giving you a porter config that looks like this:
`
default-storage: in-cluster-mongodb
default-storage-plugin: null
default-secrets-plugin: kubernetes.secrets
storage:
    - name: in-cluster-mongodb
      plugin: mongodb
`

api/v1/porterconfig_types_test.go
sgettys: Should I just omit this test case then?

carolynvs: Yeah I think so. The other cases (specifying a plugin without config), are both real use cases that cover the bug you reported. 👍

api/v1/porterconfig_types_test.go
carolynvs: Sorry I didn't catch on from the slack conversation what was wrong! 🤦‍♀️  Yes, the right thing to do is add omitempty for the other formats as well.

sgettys: Ok sounds good, I'll add that. Also DefaultSecretsPlugin doesn't have omitempty for any of the formats, I assume that it should be added to that as well for all

api/v1/porterconfig_types_test.go
sgettys: I'll make this a valid config as well

PR #88, opened 2022-03-23
Scan the manager image with Trivy during CI

carolynvs:
Scan the porter-operator-manager image with [Trivy](https://github.com/aquasecurity/trivy) for pushes and pull requests and then report the results back to GitHub Code Scanning service.

Reports for the main branch are displayed in the Security tab of the repository under "Code Scanning Alerts". For pull requests, they are displayed inline on the files tab.


magefile.go
```diff
@@ -21,6 +21,7 @@ import (
 	. "get.porter.sh/porter/mage/docker"
 	"get.porter.sh/porter/mage/releases"
 	"get.porter.sh/porter/pkg/cnab"
+	"github.com/carolynvs/magex/ci"
```
sgettys, line 24: Just a general question what's the determining factor on if something goes here versus in the getporter/magefiles repo? Should we be moving some of this code into that repo?

VinozzZ (approved):

PR #89, opened 2022-03-25
Bump porter and k8s dependencies

carolynvs:
I had to update one of our tests because the k8s test utilities that we use for our reconciler unit tests has changed its behavior. Now when a finalizer has been removed and a delete timestamp is set, the test utility removes the resource for you (just like k8s does in real life).

So yay!


vdice (approved):

bdegeeter (approved):
lgtm

also ran tests locally to confirm

PR #90, opened 2022-03-25
Make test wait time configurable

carolynvs:
The ginkgo tests often have to trigger the controller with a change, like editing an installation resource, and then wait for the controller to react and the resulting porter action jobs to finish.

Depending on if you already have the images pulled, and how fast your machine is, this can take a pretty variable amount of time. I've made the wait time configurable so that we can increase it on our CI build.

### DIFF
--- a/tests/integration/installation_test.go
+++ b/tests/integration/installation_test.go
@@ -147,9 +147,12 @@ func debugFailedInstallation(ctx context.Context, inst *porterv1.Installation) {
 func getWaitTimeout() time.Duration {
 	if value := os.Getenv("PORTER_TEST_WAIT_TIMEOUT"); value != "" {
 		timeout, err := time.ParseDuration(value)
-		if err == nil {
-			return timeout
+		if err != nil {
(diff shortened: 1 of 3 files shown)


tests/integration/installation_test.go
```diff
@@ -139,10 +143,21 @@ func debugFailedInstallation(ctx context.Context, inst *porterv1.Installation) {
...
 }
 
+// Get the amount of time that we should wait for a test action to be processed.
+func getWaitTimeout() time.Duration {
+	if value := os.Getenv("PORTER_TEST_WAIT_TIMEOUT"); value != "" {
+		timeout, err := time.ParseDuration(value)
+		if err == nil {
+			return timeout
+		}
+	}
```
VinozzZ, line 153: Maybe we can add a log or something to indicate the provided value isn't being used here due to parsing error?

carolynvs: Fixed!

VinozzZ (approved):

PR #98, opened 2022-04-27
feat: initial CredentialSet CRD/Controller

bdegeeter:
Provides for basic CredentialSet resource supporting Kubernetes secret as a credential source


bdegeeter:
We'll follow up with a docs update before taking this out of draft.

docs/content/file-formats.md
```diff
@@ -30,6 +30,44 @@ In addition to the normal fields available on a [Porter Installation document](/
...
+        secret: test-secret
+```
+
+| Field        | Required | Default                            | Description                                                 |
+|--------------|----------|------------------------------------|-------------------------------------------------------------|
+| credentials  | true     |                                    | List of credential sources for the set |
+| agentConfig  | false    | See [Agent Config](#agentconfig)   | Reference to an AgentConfig resource in the same namespace. |
+| porterConfig | false    | See [Porter Config](#porterconfig) | Reference to a PorterConfig resource in the same namespace. |
+
+### Credential Source
```
carolynvs, line 58: Can you merge this table into the Credential Set section? This page should document each individual kubernetes manifest that you can use with the Porter operator, and Credential Source isn't a standalone resource.

Here's an example of how to document nested types from the Porter Credential Set docs.

https://release-v1.porter.sh/reference/file-formats/#credential-set

docs/content/file-formats.md
```diff
@@ -30,6 +30,44 @@ In addition to the normal fields available on a [Porter Installation document](/
 
 [Installation]: /operator/glossary/#installation
 
+## CredentialSet
```
carolynvs, line 33: Can you add this to the table of contents at the top of this file?

docs/content/glossary.md
```diff
@@ -39,6 +40,18 @@ The [Installation] custom resource represents an installation of a bundle in Por
...
 [Installation]: /operator/file-formats/#installation
 
+### CredentialSet
+
+The [CredentialSet] custom resource represents a credential set in Porter.
+
+[CredentialSet]: /operator/file-formats/#credentialset
+
+A CredentialSet supports a credential source of `secret` for Porter secrets plugins. Secrets 
+must be configured based on the secret [plugin type](/plugins).
```
carolynvs, line 50:
```suggestion
A CredentialSet supports a credential source of `secret` for Porter secrets plugins. Secrets 
source keys may vary depending on which [secret plugin](/plugins/) you have configured. The [host secrets plugin](/plugins/host/) is not a good fit for use with the Porter Operator because environment variables, files are not a recommended way to manage secrets on a cluster. The [kubernetes.secrets plugin](https://release-v1.porter.sh/plugins/kubernetes/#secrets) can retrieve secrets from native Kubernetes secrets, and otherwise we recommend that an external secret store such as [Azure KeyVault](/plugins/azure/#secrets) or [Hashicorp Vault](/plugins/hashicorp/) are configured instead.
```

go.mod
```diff
@@ -2,12 +2,12 @@ module get.porter.sh/operator
...
 
-// Any replace statements should be kept in sync with the ones in Porter's go.mod
+// Any replace  statements should be kept in sync with the ones in Porter's go.mod
 
 require (
-	get.porter.sh/magefiles v0.1.3
-	get.porter.sh/porter v1.0.0-alpha.16
-	github.com/carolynvs/magex v0.7.2
+	get.porter.sh/magefiles v0.2.0
+	get.porter.sh/porter v1.0.0-alpha.13
```
carolynvs, line 9: Did you intend to drop to a lower version of Porter here?

installer/Dockerfile.tmpl
```diff
@@ -1,4 +1,4 @@
-# syntax=docker.io/docker/dockerfile:1.4
+# syntax=docker/dockerfile:1.4
```
carolynvs, line 1: Did you intend to make this change? I don't see the problem with the original line.

magefiles/magefile.go
```diff
@@ -31,9 +31,12 @@ import (
 	"github.com/carolynvs/magex/pkg/downloads"
 	"github.com/carolynvs/magex/pkg/gopath"
 	"github.com/carolynvs/magex/shx"
+	"github.com/carolynvs/magex/xplat"
 	"github.com/magefile/mage/mg"
+	"github.com/magefile/mage/target"
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
+	// mage:import
```
carolynvs, line 39: Careful running import tools on the magefiles, that comment has meaning when placed above an import line which has now been lost. In this case since the tests package is being imported with `.`, the comment isn't needed so can you just remove this stray comment now?

magefiles/magefile.go
```diff
@@ -136,6 +152,44 @@ func BuildImages() {
 	mgx.Must(p.SetEnv("MANAGER_IMAGE", img))
 }
 
+func getPlugins() error {
+	// TODO: move this to a shared target in porter
```
carolynvs, line 156: Yeah this would work best in the magefiles repository 👍

magefiles/magefile.go
```diff
@@ -245,9 +301,15 @@ func UpdateTestfiles() {
 	TestUnit()
 }
 
+func TestOutline() {
```
carolynvs, line 304: What does this target do?

magefiles/magefile.go
```diff
@@ -394,26 +468,46 @@ func namespaceExists(name string) bool {
 
 // Create a namespace, usage: mage SetupNamespace demo.
 // Configures the namespace for use with the operator.
-func SetupNamespace(name string) {
+func OldSetupNamespace(name string) {
```
carolynvs, line 471: Is it safe to remove this function? Which one is still used?

magefiles/magefile.go
```diff
@@ -455,20 +549,14 @@ func removeFinalizers(namespace, name string) {
...
 	// Get the resource definition
-	resource, _ := kubectl("get", name, "-n", namespace, "-o=yaml").Output()
-
-	// Use yq to remove the finalizers
-	resource, _ = must.Command("yq", "eval", ".metadata.finalizers = null").
-		Stdin(strings.NewReader(resource)).Output()
-
-	// Update the resource
-	kubectl("apply", "-f", "-").Stdin(strings.NewReader(resource)).Run()
+	kubectl("patch", "-n", namespace, name, "-p", `[{"op": "remove", "path": "/metadata/finalizers"}]`, "--type=json").Must(false).RunV()
```
carolynvs, line 552: 💯 This is much better, thanks!

magefiles/magefile.go
```diff
@@ -481,13 +569,13 @@ func Logs() {
...
 func EnsureOperatorSDK() {
-	const version = "v1.3.0"
+	const version = "v1.19.0"
 
 	if runtime.GOOS == "windows" {
 		mgx.Must(errors.New("Sorry, OperatorSDK does not support Windows. In order to contribute to this repository, you will need to use WSL."))
 	}
 
-	url := "https://github.com/operator-framework/operator-sdk/releases/{{.VERSION}}/download/operator-sdk_{{.GOOS}}_{{.GOARCH}}"
+	url := "https://github.com/operator-framework/operator-sdk/releases/download/{{.VERSION}}/operator-sdk_{{.GOOS}}_{{.GOARCH}}"
```
carolynvs, line 578: Sorry about that, it must have never worked?

magefiles/magefile.go
```diff
@@ -567,8 +655,93 @@ func pwd() string {
 	return wd
 }
 
+// Install the specified version of porter
+func NewEnsurePorterAt(version string) {
```
carolynvs, line 659: Can you remove this implementation and use the one in the magefiles repository? https://github.com/getporter/magefiles/blob/0b4ba70c8f969d7dfdc489310ab94a0dbb06a390/porter/porter.go#L27

magefiles/magefile.go
```diff
@@ -567,8 +655,93 @@ func pwd() string {
...
+	}
+	if os.Getenv("PORTER_AGENT_REPOSITORY") != "" && os.Getenv("PORTER_AGENT_VERSION") != "" {
+		localAgentImgName = fmt.Sprintf("%s:%s", os.Getenv("PORTER_AGENT_REPOSITORY"), os.Getenv("PORTER_AGENT_VERSION"))
+	}
+	err := buildImage(localAgentImgName)
+	mgx.Must(err)
+}
+
+//generatedCodeFilter remove generated code files from coverage report
+func generatedCodeFilter(filename string) error {
```
carolynvs, line 727: I was wondering how you so dramatically increased code coverage in this patch! 😂

magefiles/magefile.go
```diff
@@ -567,8 +655,93 @@ func pwd() string {
...
+	var lines []string
+	sc := bufio.NewScanner(strings.NewReader(fs))
+	for sc.Scan() {
+		if !strings.Contains(sc.Text(), "zz_generated") {
+			lines = append(lines, sc.Text())
+		}
+	}
+
+	fd = []byte(strings.Join(lines, "\n"))
+	err = ioutil.WriteFile(filename, fd, 0o600)
```
carolynvs, line 742: What does the o mean in this permission mode? Did you mean `0600`?

tests/integration/credset_test.go
```diff
@@ -0,0 +1,516 @@
...
+
+				// Validate that the correct credential set was used by the installation
+				jsonOut := runAgentAction(ctx, "show-outputs", ns, []string{"installation", "outputs", "list", "-n", ns, "-i", inst.Spec.Name, "-o", "json"})
+				credsValue := gjson.Get(jsonOut, `#(name=="outInsecureValue").value`).String()
+				Expect(credsValue).To(Equal(testSecret))
+			})
+		})
+	})
+})
+var _ = Describe("CredentialSet secret does not exist", func() {
```
carolynvs, line 75: Do you know what the overhead/implications are of splitting these up into separate describe sections vs multiple checks in the same one? I'm wondering if we setup/tearndown just the test namespace for each of these, or if it also redoes the envtest cluster too?

tests/integration/credset_test.go
```diff
@@ -0,0 +1,516 @@
...
+			}
+
+			time.Sleep(time.Second)
+			continue
+		}
+	}
+
+}
+
+func waitForPorterCS(ctx context.Context, cs *porterv1.CredentialSet, msg string) error {
```
carolynvs, line 354: I wonder if some of these helper functions can be made more generic with an Object interface instead or any other tricks so that we don't have to duplicate these functions for each type? 🤔 

If you figure that out, it would make a great follow-up PR.

tests/integration/credset_test.go
```diff
@@ -0,0 +1,516 @@
...
+				err := waitForPorter(ctx, inst, "waiting for porter-test-me to install")
+				Expect(err).Should(HaveOccurred())
+				validateInstallationConditions(inst)
+				Expect(inst.Status.Phase).To(Equal(porterv1.PhaseFailed))
+			})
+		})
+	})
+})
+
+// TODO: Add failed installation due to missing secret test
```
carolynvs, line 116: The todo here is referencing functionality in porter, not the operator, since as we discussed at the last meeting, the operator isn't aware of or validating what exists before calling porter.

Maybe switch this todo to an issue and describe the functionality you'd like to see in the operator here.

tests/integration/credset_test.go
```diff
@@ -0,0 +1,516 @@
...
+				Expect("test-credential").To(Equal(updatedFirstCredName))
+				Expect("new-credential").To(Equal(updatedSecondCredName))
+			})
+		})
+	})
+})
+
+var _ = Describe("CredentialSet delete", func() {
+	Context("when an existing CredentialSet is delete", func() {
+		It("should run porter credentials delete", func() {
```
carolynvs, line 188: This test seems to be a copy of the update test above, with the extra delete action and versification. One way to make this easier to maintain going forward would be to combine them into a single test that walks through the entire lifecycle of a credential set.

tests/integration/credset_test.go
```diff
@@ -0,0 +1,516 @@
...
+	cs := &porterv1.CredentialSet{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "porter.sh/v1",
+			Kind:       "CredentialSet",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "porter-test-me-",
+		},
+		Spec: porterv1.CredentialSetSpec{
+			//TODO: get schema version from porter version?
```
carolynvs, line 255: yeah you could totally read the schema version from get.porter.sh/porter/pkg/credentials.SchemaVersion. I just created a PR that moves this value around a bit (https://github.com/getporter/porter/pull/2052) but I like how this would simplify and reduce some of the hard coding in our test data.

tests/integration/suite_test.go
```diff
@@ -112,20 +116,21 @@ var _ = AfterSuite(func() {
 })
 
 var _ = BeforeEach(func() {
-	testNamespace = createTestNamespace(context.Background())
+	schemaVersion = "1.0.1"
```
carolynvs, line 119: This is a bit confusing because the schema version for installations, credsets and param sets can all be different (though I think the stars aligned recently and they are all the same for now).

tests/integration/suite_test.go
```diff
@@ -112,20 +116,21 @@ var _ = AfterSuite(func() {
...
 }, 5)
 
 var _ = AfterEach(func() {
 	if _, ok := os.LookupEnv("KEEP_TESTS"); ok {
 		return
 	}
-	deleteNamespace(testNamespace)
+	//TODO: don't reuse namespace for all tests move this to suite/magefile cleanup
+	//deleteNamespace(testNamespace)
```
carolynvs, line 127: What was the problem with the original code that made you comment it out? Let's get this TODO addressed in this PR so that we don't lose track of it.

tests/integration/testdata/operator_porter_config.yaml
```diff
@@ -0,0 +1,13 @@
+#TODO: remove this file
```
carolynvs, line 1: Can you either remove this file or the confusing comment 😁

carolynvs (changes requested):
One more that I couldn't comment directly on the file (kubernernetes-linux-amd64) should not be checked in. If you need the plugin binary here, let's grab it using the magefile. Or if you have already done that, this should be in the .gitignore so that it's not accidentally checked in?

.github/workflows/build.yml
```diff
@@ -20,10 +24,6 @@ jobs:
         with:
           go-version: 1.18
           check-latest: true
-      - name: Setup Porter
```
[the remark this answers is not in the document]
bdegeeter: Thanks for catching the binary check in.  This was from before there was one published we could download.  I'll add it to the magefile.

api/v1/credentialset_types.go
```diff
@@ -0,0 +1,133 @@
...
+	Source CredentialSource `json:"source" yaml:"source"`
+}
+
+// CredentialSource defines a element in a CredentialSet
+type CredentialSource struct {
+	//Secret is a credential source using a secret plugin
+	Secret string `json:"secret,omitempty" yaml:"secret,omitempty"`
+
+	//FilePath is a credential source using a file path (and configmap)
+	FilePath string `json:"file,omitempty" yaml:"file,omitempty"`
```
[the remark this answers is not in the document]
carolynvs: Sounds like we don't actually support path/file yet? Let's remove this field and add it back when we are ready to support it.

tests/integration/suite_test.go
bdegeeter: The goal was to make test execution and cleanup more resilient.  Namespace per test was cleaner during feature development.  It also helps support parallel execution of tests.  The commented out code and TODO should have been removed.  We could also remove the KEEP_TEST flag at this point, because the cleanup always happens at the start of the next run.  Let me know what you think.

carolynvs: Does each test get a new namespace to run under? I'm wondering how you are running tests in parallel and also cleaning up previous test runs if they are generated names?

api/v1/credentialset_types.go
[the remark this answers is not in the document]
bdegeeter: Just noticed that myself :)  I'm removing it.

magefiles/magefile.go
bdegeeter: It creates a summary of behaviors we'd like to work on for BDD.  https://onsi.github.io/ginkgo/#creating-an-outline-of-specs

tests/integration/suite_test.go
bdegeeter: Yes, each test gets its own namespace (still labeling with `porter.sh/testdata`).  The TestIntegration target removes all namespaces with the label at the start of a run.

magefiles/magefile.go
bdegeeter: I think this was a recent change.  It worked before.

magefiles/magefile.go
bdegeeter: I hope you don't think it's cheating  :)

magefiles/magefile.go
bdegeeter: changed to `0600`.  I think this was a darwin thing.

tests/integration/credset_test.go
bdegeeter: agreed, we should follow up to remove this dup

api/v1/credentialset_types_test.go
```diff
@@ -53,12 +54,11 @@ func TestCredentialSetSpec_ToPorterDocument(t *testing.T) {
 				Credentials:   tt.fields.Credentials,
 			}
 			got, err := cs.ToPorterDocument()
-			if (err != nil) != tt.wantErr {
-				t.Errorf("CredentialSetSpec.ToPorterDocument() error = %v, wantErr %v", err, tt.wantErr)
-				return
-			}
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("CredentialSetSpec.ToPorterDocument() = \n%v, want \n%v", string(got), string(tt.want))
+			if tt.wantErrMsg == "" {
```
carolynvs, line 57: It looks like you are missing an else statement here to handle when wantErrMsg is not empty?

carolynvs (changes requested):
This looks great! If you can just fix that one test, this is ready to go. 👍

carolynvs (approved):
oops sorry I see that you do have that else statement, I just wasn't seeing it thanks to the magic of contextual diffs. :shipit:

PR #103, opened 2022-06-09
Integration test helpers rewrite

sgettys:
* Update the integration tests to use the published agent image that includes the k8s plugin
* Updated to 1.0-alpha.20
* Consolidated integration test helpers using dynamic client. The reason for this is due to different concrete types that's used for the agent action status versus the porter resources status. Looked into different ways to solve this and dynamic client seems to be cleanest option without adding unnecessary interfaces or multiple abstraction layers


tests/integration/installation_test.go
```diff
@@ -55,143 +47,31 @@ var _ = Describe("Installation Lifecycle", func() {
...
-			Expect(waitForPorter(ctx, inst, "waiting for the bundle to install")).Should(Succeed())
-			validateInstallationConditions(inst)
+			Expect(waitForPorter(ctx, inst, 1, "waiting for the bundle to install")).Should(Succeed())
+			validateResourceConditions(inst.Status.Conditions)
 
 			patchInstallation := func(inst *porterv1.Installation) {
 				controllers.PatchObjectWithRetry(ctx, logr.Discard(), k8sClient, k8sClient.Patch, inst, func() client.Object {
 					return &porterv1.Installation{}
 				})
+				// Wait for patch to apply, this can cause race conditions
```
carolynvs, line 57: I see a comment here but no corresponding code below. Does this comment apply to something I'm just missing or did you intend to add more waiting here?

tests/integration/utils_test.go
```diff
@@ -0,0 +1,298 @@
...
+	Expect(apimeta.IsStatusConditionTrue(conditions, string(porterv1.ConditionStarted)))
+	Expect(apimeta.IsStatusConditionTrue(conditions, string(porterv1.ConditionComplete)))
+}
+
+// Get the pod logs associated to the job created by the agent action
+func getAgentActionJobOutput(ctx context.Context, agentActionName string, namespace string) (string, error) {
+	actionKey := client.ObjectKey{Name: agentActionName, Namespace: namespace}
+	action := &porterv1.AgentAction{}
+	if err := k8sClient.Get(ctx, actionKey, action); err != nil {
+		Log(errors.Wrap(err, "could not retrieve the CredentialSet's AgentAction to troubleshoot").Error())
```
carolynvs, line 233: Same as above.

carolynvs (commented):
YES! This is so much better. Thanks for de-duping all that code.

tests/integration/installation_test.go
sgettys: Whoops this was before I figured out the race conditions in the waitForPorter, will remove

carolynvs (approved):
Looks great, thanks for cleaning up our copy/paste sins!

Issue #54, opened 2021-11-08
Delete temporary resources when agent completes


carolynvs:
When the agent completes, delete any temporary resources created by the operator such as: jobs, pvcs, secrets.

carolynvs:
See https://sdk.operatorframework.io/docs/best-practices/resource-pruning/

carolynvs:
It would be nice to have some configuration options:

* feature flag to disable pruning
* TTL or some other way to say how long to keep them before deleting
* Possibly keep failed runs longer or treat differently than passing jobs

carolynvs:
The plan is to add `TTLSecondsAfterFinished` to AgentConfig. This controls how an AgentAction and associated resources (job, pv, pvc, secrets, etc) should remain after the action finishes (success or failure).

Resources created by the operator
* Agent Job
* Agent PVC
* Agent PV (share bundle outputs and files between porter and the bundle)
* Porter Config Secret
* Porter WorkDir Secret

Resources created by the agent (k8s driver)
* Bundle Job
* Bundle Secret (inject env vars into bundle)

I have created https://github.com/cnabio/cnab-go/issues/300 to track supporting owner references in the k8s driver

carolynvs:
I chose to not go with the operator framework pruning library because it would be on us to schedule and call it regularly. There aren't any documented best practices on how to run that in an operator at this time and I didn't find any use of the library in existing projects to look for ideas.

The simplicity of pawning off the pruning trigger onto k8s instead seems more reliable to me.

PR #108, opened 2022-07-12
Add ability to install porter plugins

VinozzZ:
This PR implements the ability to allow users to configure plugins they need in agent-config resource.

Current workflow:

It  creates a new agent config controller to manage the lifecycle of the plugin installation. The controller dispatch jobs to the porter agent to install plugins onto a persistent volume through a pvc (temporary persistent volume claim) managed by the agent config controller. Once the installation is done, the controller sets labels, including all installed plugins and the agent config name,  on the persistent volume. Then the controller will create a new pvc.  The new pvc will use the hash of all the plugins information as its name and the plain text version of all the plugin information as its label.
When a new agent action is created, it will query using the plugin hash derived from its agent config and mount a plugin pvc using that name.
If no existing plugin pvc exits, the agent action will wait for the pvc to be created before kubernetes schedules it.
When a plugin volumes is no longer needed, the current approach is to allow k8s GC to do the work for us. All the operator is doing when a agent config is deleted  is to make sure to remove itself from the pvc's and pv's `OwnerReferences` value. Once no other resource is referencing a volume, GC will know to delete it permanently.
To make sure we are not deleting any volumes that still could being used, when a volume is being reused by multiple agent config, the controller will make sure to add all agent config into the owner reference before marking the agent config status has complete.

Couple issues that may need to be worked out before this feature is complete:

- [x] - Since currently, the operator doesn't have the capability to clean up pods after each job has been finished, the above cleanup workflow for plugin volumes only removes the deleted agent config from the plugin volume's owner reference list. we need to validate that when all the pods are deleted, the plugin volumes will be deleted too. https://github.com/getporter/operator/issues/54

    **Resolution: I would like to work on this as a separate PR and just have the current deletion behavior as only removing `OwnerReference`.**

- [x] - Another issue that the current code has not solved is how to install multiple porter plugins with one agent job. From my discovery, porter currently does not seem to allow installation for multiple plugins with one command. https://github.com/getporter/operator/issues/123

   **Resolution: This will be a separate PR in porter to implement this new feature. For now, only `kubernetes` plugins will be installed. Users will only able to define `kubernetes` plugin and configure its version and feedURL**

- [x] - We probably should set default plugin configuration in both the installer bundle and the controller code. If someone were to delete the default agent config defined by the installer bundle, the controller can still set a default kubernetes pluin in the code
- [x] - We need to add integration tests for delete behavior
- [x] - This feature is also missing document. For example, where all the default values for plugin configs are set https://github.com/getporter/operator/issues/122


carolynvs:
BTW I have codecov turned on mostly for our own information. It's not a requirement to hit a certain percentage and it's okay if a feature reduces coverage (which is easy to do since we don't test exception handling).

api/v1/agentaction_types.go
```diff
@@ -78,6 +78,16 @@ func (a *AgentAction) GetRetryLabelValue() string {
 	return getRetryLabelValue(a.Annotations)
 }
 
+func (a *AgentAction) IsAgentConfig() bool {
+	for _, ref := range a.GetOwnerReferences() {
+		if ref.Kind == "AgentConfig" {
```
sgettys, line 83: nit: Define in agentconfig_types.go instead of hardcoded string

api/v1/agentconfig_types.go
```diff
@@ -91,6 +98,55 @@ func (c AgentConfigSpec) GetVolumeSize() resource.Quantity {
...
+		plugins = append(plugins, p.Name)
+
+		if p.Version != "" {
+			plugins = append(plugins, fmt.Sprintf("_%s", p.Version))
+		}
+	}
+
+	return map[string]string{
+		LabelManaged: "true",
+		LablePlugins: strings.Join(plugins, ""),
```
sgettys, line 146: nit: "Label"

api/v1/const.go
```diff
@@ -40,6 +40,8 @@ const (
 	// Operator.
 	LabelManaged = Prefix + "managed"
 
+	LablePlugins = Prefix + "plugins"
```
sgettys, line 43: nit: Lable -> Label throughout

docs/content/quickstart/_index.md
```diff
@@ -312,6 +312,31 @@ Currently the Operator only supports the first imagePullSecret in a service acco
...
+  installationServiceAccount: installation-agent
+  plugins:
+    - name: "kubernetes"
+```
+
+Create the `AgentConfig` custom resource by running `kubectl apply -f quickstart_agentconfig.yaml`
+
+If no plugins are defined in a AgentConfig, the operator will set its default value to use the `kubernetes-plugin`
+
+🚨 WARNING: Currently, the operator can only install one plugin per AgentConfig. If more than one plugins are defined in the CRD, it will omit the rest of the defined plugins but the first one.
```
sgettys, line 338: Is this because of the limitations of calling "porter" inside the agent action and installing multiple plugins would currently require multiple porter commands? Having the ability to run multiple porter commands inside a single agent action came up before when we were working on some outputs stuff and it might be a nice addition in general to have to the agent image. It would be a simple update to https://github.com/getporter/porter/blob/main/cmd/agent/main.go to add that in and we could get that behavior in addition to potentially allowing this for other use cases. I'm also fine with leaving it as is and just waiting until porter directly supports installing multiple plugins via a single plugins install command

controllers/agentconfig_controller.go
```diff
@@ -0,0 +1,706 @@
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if processed {
+		err = removeAgentCfgFinalizer(ctx, log, r.Client, agentCfg)
+		log.V(Log4Debug).Info("Reconciliation complete: Finalizer has been removed from the AgentConfig.")
+		return ctrl.Result{}, err
+	}
+
+	// TODO: once porter has ability to install multiple plugins with one command, we will allow users
```
sgettys, line 96: Couldn't we at least add the ability to specify a single plugin now or is this just going to be a WIP feature until porter is ready?

controllers/agentconfig_controller.go
```diff
@@ -0,0 +1,706 @@
...
+		}
+
+		// if the plugin install action is not finished, we need to wait for it before acting further
+		if !(apimeta.IsStatusConditionTrue(action.Status.Conditions, string(porterv1.ConditionComplete)) && action.Status.Phase == porterv1.PhaseSucceeded) {
+			log.V(Log4Debug).Info("Plugins is not ready yet.", "action status", action.Status)
+			return ctrl.Result{}, nil
+		}
+
+		// delete the temporary pvc first once the plugin install action has been completed
+		shouldDeleteTmpPVC := tempPVC != nil && readyPVC == nil
```
sgettys, line 152: nit: Just directly check `if tempPVC != nil && readyPVC == nil`

controllers/agentconfig_controller.go
```diff
@@ -0,0 +1,706 @@
...
+	if len(results.Items) == 0 {
+		log.V(Log4Debug).Info("No existing agent action was found")
+		return nil, false, nil
+	}
+	action := results.Items[0]
+	log.V(Log4Debug).Info("Found existing agent action", "agentaction", action.Name)
+	return &action, true, nil
+}
+
+// Run the porter agent with the command `porter agent config apply`
```
sgettys, line 239: Update comment, runs "porter plugins install"

controllers/agentconfig_controller.go
```diff
@@ -0,0 +1,706 @@
...
+	action.SetRetryAnnotation(retry)
+	if err := r.Update(ctx, action); err != nil {
+		return errors.Wrap(err, "error updating the associated porter agent action")
+	}
+
+	log.V(Log4Debug).Info("Retried associated porter agent action", "name", "retry", action.Name, retry)
+	return nil
+}
+
+func getPluginVolumn(pvc *corev1.PersistentVolumeClaim) (corev1.Volume, corev1.VolumeMount) {
```
sgettys, line 406: nit: getPluginVolume

controllers/agentconfig_controller.go
```diff
@@ -0,0 +1,706 @@
...
+	pvKey := client.ObjectKey{Namespace: newPVC.Namespace, Name: newPVC.Spec.VolumeName}
+	err = r.Get(ctx, pvKey, pv)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	if idx, exist := containOwner(pv.GetOwnerReferences(), agentCfg); exist {
```
sgettys, line 511: Add comment here and below on what's different between these two calls to removeOwnerByIdx

controllers/agentaction_controller.go
```diff
@@ -713,14 +711,27 @@ func (r *AgentActionReconciler) getAgentVolumes(action *porterv1.AgentAction, pv
 		},
 		)
 	}
-
-	for _, volume := range action.Spec.Volumes {
-		volumes = append(volumes, volume)
+	if !action.IsAgentConfig() {
+		claimName := agentCfg.GetPVCName(action.Namespace)
```
sgettys, line 715: What if the AgentConfig has a non Succeeded status? Does an AgentAction that is using an AgentConfig that is still being created block until that AgentConfig job finishes? If an AgentAction is using an AgentConfig that failed to install the plugins will that AgentAction fail as well?

controllers/agentaction_controller.go
```diff
@@ -713,14 +711,27 @@ func (r *AgentActionReconciler) getAgentVolumes(action *porterv1.AgentAction, pv
 		},
 		)
 	}
-
-	for _, volume := range action.Spec.Volumes {
-		volumes = append(volumes, volume)
+	if !action.IsAgentConfig() {
```
carolynvs, line 714: Can you add a comment here? It's not immediately clear what's going on and an explanation would help a lot. I can't really tell from the code when we mount this volume.

controllers/agentconfig_controller.go
carolynvs: Agreed, using the config from the AgentConfig CRD, and just acknowledging that we only install the first one would be preferable.

controllers/agentconfig_controller.go
```diff
@@ -0,0 +1,702 @@
...
+	if updatedCfg {
+		err := r.Update(ctx, agentCfg)
+		return ctrl.Result{}, err
+	}
+
+	readyPVC, tempPVC, err := r.GetPersistentVolumeClaims(ctx, log, agentCfg)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if readyPVC != nil && tempPVC == nil && !isDeleted(agentCfg) {
```
carolynvs, line 108: I recommend moving all of this into multiple functions. If you look over the other reconcile functions in the other controller, I've tried real hard to keep them short and readable because otherwise this function is overwhelming and difficult to test/maintain.

carolynvs (changes requested):
I didn't have time to do a full review but here is some initial feedback. I can take another review after the logic in reconcile is split up into functions.

api/v1/agentconfig_types.go
```diff
@@ -91,6 +98,55 @@ func (c AgentConfigSpec) GetVolumeSize() resource.Quantity {
...
+	}
+
+	var plugins []string
+	for _, p := range c.Plugins {
+		if p.Name == "" {
+			continue
+		}
+		plugins = append(plugins, p.Name)
+
+		if p.Version != "" {
```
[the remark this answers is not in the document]
VinozzZ: I thought so since the `porter plugins install kubernetes` has the default behavior to install the latest release version

controllers/agentconfig_controller.go
VinozzZ: Yes, the comment was not correct. I have updated it

controllers/agentaction_controller.go
VinozzZ: Yes, the AgentAction is blocked until the AgentConfig finishes installing plugins and the plugin volume is ready.
If an AgentConfig fails, the plugin volume claim would not exist, I would assume the following agent action would timeout eventuall(I tried to google information on the timeout but I haven't able to confirm the timeout behavior).
Do you have any suggestion on how to we can better handle the failure case here? To me, if the plugins are not ready, there's probably some serious issue going on and the operator should not try to continue working.

api/v1/agentconfig_types.go
```diff
@@ -139,3 +234,10 @@ type AgentConfigList struct {
 func init() {
 	SchemeBuilder.Register(&AgentConfig{}, &AgentConfigList{})
 }
+
+type PluginList []Plugin
+type Plugin struct {
+	Name    string `json:"name,omitempty" mapstructure:"name,omitempty"`
```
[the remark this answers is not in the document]
VinozzZ: Oh, I did not know this. Thank you for mentioning it. 
I think I did the checks in the code is mostly due to issues I ran into with tests. Do you know where in the codebase I can find the logic that all fields are required by default?

controllers/agentaction_controller.go
sgettys: I think the action should fail if the AgentConfig associated with that action didn't successfully install the plugins. Just wanted to make sure that was how this worked I didn't see how that was happening in the code currently. We could also check the associated status of the AgentConfig resource we get with the AgentAction and if its in "Failed" state then fail the AgentAction. I'm not sure on the k8s internal behavior though if you try to bind a PVC that doesn't exist to a job will it wait for that to exist before attempting the job or will it fail immediately?

This is the behavior I would expect:
* Given: AgentConfig is still running
   * When: An AgentAction is created that uses that AgentConfig
      * Then: The AgentAction will wait until the AgentConfig reaches a finished state before creating the porter job for that AgentAction
* Given: AgentConfig resource is running
   * When: 
      * Then: Any waiting AgentActions relying on that AgentConfig will fail
      * And: Any new AgentActions that reference that AgentConfig will fail
* Given: No AgentConfig
   * When: A new AgentAction is created
      * Then: A default AgentConfig is created
      * And: The AgentAction waits for that AgentConfig to reach a completed state

Do those sound reasonable and is this currently what the behavior is?

VinozzZ: Thanks for the clarification. In your list of the expected behaviors, do you mind to explain what's the difference between the first and the second scenarios? 
Also I'm not sure an AgentAction has an associated AgentConfig once all the configs(AgentConfig and porter configs) are resolved. On this line https://github.com/getporter/operator/blob/5ac5ff17a3473c663eb1f24cc19305b75858188c/controllers/agentaction_controller.go#L468, the `resolveAgentConfig` function merges default values with the AgentConfig defined on the AgentAction. The plugins could be installed by a namespace level AgentConfig not the one defined on the AgentAction.
To answer your question above, k8s will wait to attempt a job if a binded PVC doesn't exit yet and will proceed once the PVC is ready.

api/v1/agentaction_types.go
```diff
@@ -78,9 +78,10 @@ func (a *AgentAction) GetRetryLabelValue() string {
 	return getRetryLabelValue(a.Annotations)
 }
 
-func (a *AgentAction) IsAgentConfig() bool {
+// CreatedByAgentConfig checks if an AgentAction is running on behalf of an agent config.
```
carolynvs, line 81: Thanks this is more clear to me! 👍

controllers/agentconfig_controller.go
```diff
@@ -0,0 +1,730 @@
...
+	if processed {
+		err = removeAgentCfgFinalizer(ctx, log, r.Client, agentCfg)
+		log.V(Log4Debug).Info("Reconciliation complete: Finalizer has been removed from the AgentConfig.")
+		return ctrl.Result{}, err
+	}
+
+	// TODO: once porter has ability to install multiple plugins with one command, we will allow users
+	// to install multiple plugins. Currently, only the first item defined in the plugin list will be
+	// installed.
+	updatedCfg := setDefaultPlugins(agentCfg)
```
carolynvs, line 99: re-adding comment from last review since I think it was missed:

I recommend moving most of the detailed logic from Reconcile into multiple functions. If you look over the other reconcile implementations in the other controller, I've tried real hard to keep them short/readable, sticking to the high level status transitions, because otherwise this function is overwhelming and difficult to test/maintain.

controllers/agentaction_controller.go
sgettys: Whoops sorry that was supposed to be "When: The AgentConfig AgentAction fails".

So there's no way for an AgentAction to resolve to a specific AgentConfig based on what is specified in its AgentConfig, that's just a reference to the AgentConfigSpec not an actual resource correct? That makes sense so it would be tough to know if that corresponding AgentAction succeeded. I just don't want to get into a point where an Installation can hang either indefinitely or for a very long time waiting for a PVC that will never appear

docs/content/quickstart/_index.md
VinozzZ: Yes, you are right on point. That's definitely something that worth looking into adding. However, I didn't dig too deep into the design for making that change. I have created an issue to track this https://github.com/getporter/porter/issues/2492 
If you want to add more detailed design in the issue, please feel free to!

controllers/agentaction_controller.go
VinozzZ: Hey, I have added a new status field on AgentConfig resource to indicate whether an AgentConfig is ready(meaning the plugins has been successfully installed). Any agent action job will only be created once the AgentConfig is ready. If it's not ready, the agent action reconcile loop will do the retry

controllers/agentaction_controller.go
```diff
@@ -476,31 +476,41 @@ func (r *AgentActionReconciler) resolveAgentConfig(ctx context.Context, log logr
...
 
+	var isCfgReady bool
 	// Read agent configuration defined at the system level
 	systemCfg := &porterv1.AgentConfig{}
 	err := r.Get(ctx, types.NamespacedName{Name: "default", Namespace: operatorNamespace}, systemCfg)
 	if err != nil && !apierrors.IsNotFound(err) {
-		return porterv1.AgentConfigSpec{}, errors.Wrap(err, "cannot retrieve system level porter agent configuration")
+		return porterv1.AgentConfigSpecAdapter{}, errors.Wrap(err, "cannot retrieve system level porter agent configuration")
 	}
+	isCfgReady = err == nil && systemCfg.Status.Ready
```
sgettys, line 490: Can you wrap this in () to make it more clear what the assignment is?

controllers/agentaction_controller.go
```diff
@@ -476,31 +476,41 @@ func (r *AgentActionReconciler) resolveAgentConfig(ctx context.Context, log logr
...
 	if action.Spec.AgentConfig != nil {
 		err = r.Get(ctx, types.NamespacedName{Name: action.Spec.AgentConfig.Name, Namespace: action.Namespace}, instCfg)
-		if err != nil && !apierrors.IsNotFound(err) {
-			return porterv1.AgentConfigSpec{}, errors.Wrapf(err, "cannot retrieve agent configuration %s specified by the agent action", action.Spec.AgentConfig.Name)
+		if !apierrors.IsNotFound(err) {
+			if err != nil {
+				return porterv1.AgentConfigSpecAdapter{}, errors.Wrapf(err, "cannot retrieve agent configuration %s specified by the agent action", action.Spec.AgentConfig.Name)
+
+			}
+			isCfgReady = err == nil && instCfg.Status.Ready
```
sgettys, line 513: Same as above

controllers/agentaction_controller.go
```diff
@@ -476,31 +476,41 @@ func (r *AgentActionReconciler) resolveAgentConfig(ctx context.Context, log logr
...
-		return porterv1.AgentConfigSpec{}, errors.Wrap(err, "cannot retrieve namespace level porter agent configuration")
+	if !apierrors.IsNotFound(err) {
+		if err != nil {
+			return porterv1.AgentConfigSpecAdapter{}, errors.Wrap(err, "cannot retrieve namespace level porter agent configuration")
+		}
+		isCfgReady = nsCfg.Status.Ready
 	}
 	logConfig("namespace", nsCfg)
 
 	// Read agent configuration override
```
sgettys, line 504: Just wanted to wrap my head around this. So there's the Agent Config that's set as "default" in the namespace and those values get read in then the actual agent action could potentially be using a non default named agent config which gets merged with the default namespace config? So you can define overrides in a named config and it will merge with the default config? My question is what happens if there's a plugin defined in the default config and a different plugin in the override config? Those plugin lists will merge correct? In which case will the appropriate volumes be mounted properly? Since the default config will point to one volume for its plugins and the named config will point to another. Or does the named config just override the whole "plugins" portion of the agent config and we only use that value and disregard the plugins list in the default config?

carolynvs (commented):
Wow, this is a giant feature. I had no idea when we first started that it would require so much careful logic and design to work properly.

Thank you for taking the time to think through all the workflow and edge cases. This is amazing and I'm really excited to get it merged so everyone can use it! 💖

carolynvs (approved):
This is great! Ship it 🚀

PR #116, opened 2022-09-07
update porter to v1.0.0-rc.1

VinozzZ:
- update porter to v1.0.0-rc.1 and set it as the default porter-agent version
- use kubernetes-plugin v1.0.0-rc.1 in integration test
- update getporter/magefiles to v0.3.2
After this PR is merged, we will do release v0.7.0 for the porter-operator


magefiles/magefile.go
```diff
@@ -161,7 +157,7 @@ func getPlugins() error {
 		feed    string
 		version string
 	}{
-		{name: "kubernetes", feed: "https://cdn.porter.sh/plugins/atom.xml", version: "latest"},
```
VinozzZ, line 164: It doesn't look like we are publishing to the latest tag anymore

VinozzZ:
This PR needs to wait for the ghcr.io/getporter/dev/porter-agent-kubernetes:v1.0.0-rc.1 to be published first

magefiles/magefile.go
carolynvs: When you run `porter plugin install` and use the latest version, it downloads the atom.xml file and selects the highest version of that plugin, excluding pre-releases.

carolynvs (approved):
LGTM!

PR #119, opened 2022-09-09
Check for failed condition in credential controller test

VinozzZ:
This PR fixes a check in integration test to actually check for expected condition.
In the `credset_test.go`, we were not correctly checking for a failed installation and this PR fixes that.
When debug for a failed resource, the order of parameters were not passed in correctly, this PR fixes it.

### DIFF
--- a/tests/integration/utils_test.go
+++ b/tests/integration/utils_test.go
@@ -140,7 +140,7 @@ func waitForPorter(ctx context.Context, resource client.Object, expGeneration in
 					return nil
 				}
 				if apimeta.IsStatusConditionTrue(conditions, string(porterv1.ConditionFailed)) {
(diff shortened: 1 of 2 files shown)


tests/integration/utils_test.go
```diff
@@ -140,7 +140,7 @@ func waitForPorter(ctx context.Context, resource client.Object, expGeneration in
 					return nil
 				}
 				if apimeta.IsStatusConditionTrue(conditions, string(porterv1.ConditionFailed)) {
-					debugFailedResource(ctx, name, namespace)
+					debugFailedResource(ctx, namespace, name)
```
carolynvs, line 143: oops! 😁

carolynvs (approved):
LGTM! 👍

PR #127, opened 2022-10-21
test: Updated waitForPorter utils test helper

sgettys:
Instead of using unstructured to get the resource status fetch it directly from the resource. This should remove a race condition where the resource status is out of sync with what is in k8s when immediately checking status of a resource after applying and waiting for it to complete

Tested locally, no failures across multiple runs of integration tests


carolynvs (approved):
This is a great improvement, race condition or not. Much easier to understand! 👍

PR #136, opened 2023-01-05
run vet and lint in CI

VinozzZ:
This PR adds vet and lint stage in github action.
For linting, the PR adds staticcheck into the repo

### DIFF
--- a/api/v1/porterconfig_types.go
+++ b/api/v1/porterconfig_types.go
@@ -91,12 +91,12 @@ func MergeMap(target, override map[string]interface{}) map[string]interface{} {
 
 // SecretsConfig is the plugin stanza for secrets.
 type SecretsConfig struct {
-	PluginConfig `json:",squash" yaml:",inline" mapstructure:",squash"`
+	PluginConfig `yaml:",inline" mapstructure:",squash"`
 }
 
 // StorageConfig is the plugin stanza for storage.
 type StorageConfig struct {
-	PluginConfig `json:",squash" yaml:",inline" mapstructure:",squash"`
+	PluginConfig `yaml:",inline" mapstructure:",squash"`
 }
 
 // PluginConfig is a standardized config stanza that defines which plugin to
--- a/controllers/agentaction_controller.go
+++ b/controllers/agentaction_controller.go
@@ -165,16 +165,17 @@ func (r *AgentActionReconciler) applyJobToStatus(log logr.Logger, action *porter
 		setCondition(log, action, porterv1.ConditionStarted, "JobStarted")
 	}
 
+conditions:
 	for _, condition := range job.Status.Conditions {
 		switch condition.Type {
 		case batchv1.JobComplete:
 			action.Status.Phase = porterv1.PhaseSucceeded
 			setCondition(log, action, porterv1.ConditionComplete, "JobCompleted")
-			break
+			break conditions
 		case batchv1.JobFailed:
 			action.Status.Phase = porterv1.PhaseFailed
 			setCondition(log, action, porterv1.ConditionFailed, "JobFailed")
-			break
+			break conditions
 		}
 	}
 }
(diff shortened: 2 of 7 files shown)


api/v1/porterconfig_types.go
```diff
@@ -91,12 +91,12 @@ func MergeMap(target, override map[string]interface{}) map[string]interface{} {
 
 // SecretsConfig is the plugin stanza for secrets.
 type SecretsConfig struct {
-	PluginConfig `json:",squash" yaml:",inline" mapstructure:",squash"`
+	PluginConfig `json:",inline" yaml:",inline" mapstructure:",squash"`
```
carolynvs, line 94: This is always confusing! 😅  

I am pretty sure that the json tag does not support squash or inline. Embedded fields are always inlined and do not require a json entry at all.

I'm pretty sure that the json tag just ignores the `,inline` anyway but it may be more clear (so that people aren't tricked into thinking it is doing anything) if we remove that json tag entirely.

```suggestion
	PluginConfig `yaml:",inline" mapstructure:",squash"` # json is inline automatically because it's an embedded field
```

controllers/agentaction_controller.go
```diff
@@ -165,16 +165,17 @@ func (r *AgentActionReconciler) applyJobToStatus(log logr.Logger, action *porter
...
 	}
 
+conditions:
 	for _, condition := range job.Status.Conditions {
 		switch condition.Type {
 		case batchv1.JobComplete:
 			action.Status.Phase = porterv1.PhaseSucceeded
 			setCondition(log, action, porterv1.ConditionComplete, "JobCompleted")
-			break
+			break conditions
```
carolynvs, line 174: I'm unclear on the intent of the change and what we want the code to be doing. The way I'm reading this, when a condition on the job is complete/failed, we will call setCondition, and then repeat the entire for loop over again, re-evaluating the original condition that called the goto.

Is there a way we can do this without the goto statement (break to a label)? At the very least we need a lot more comments explaining what's intended and make sure we have a test to validate that this can't cause an infinite loop.

mage/docs/docs_test.go
```diff
@@ -12,39 +11,30 @@ import (
 
 func TestEnsurePorterRepository(t *testing.T) {
 	t.Run("has local repo", func(t *testing.T) {
-		tmp, err := ioutil.TempDir("", "porter-docs-test")
-		require.NoError(t, err)
-		defer os.RemoveAll(tmp)
+		tmp := t.TempDir()
```
carolynvs, line 14: So much cleaner! 💯

magefiles/magefile.go
```diff
@@ -93,6 +92,11 @@ func Vet() {
 	must.RunV("go", "vet", "./...")
 }
 
+func Lint() {
+	tools.EnsureStaticCheck()
```
carolynvs, line 96: I can't remember if we do it like this elsewhere. Normally, we would call EnsureStaticCheck with mg.Dep so that it's only called once even when multiple targets call Lint. Is there a reason why we aren't doing that here?

e.g.

```suggestion
	mg.Dep(tools.EnsureStaticCheck)
```

carolynvs (commented):
Looks good! I just have a few questions/nits.

api/v1/porterconfig_types.go
VinozzZ: I think you are right on the embedded fields are always inlined and do not require a json entry when used with `encoding/json`, the standard library.
However, it looks like the `controller-gen` requires it to be specified. I found the same `json:,inline; in their doc as well. https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html

magefiles/magefile.go
VinozzZ: I fixed in porter as well: https://github.com/getporter/porter/pull/2515/files

controllers/agentaction_controller.go
VinozzZ: I misunderstood the original code. I think the break statement is redundant and I just removed them

api/v1/porterconfig_types.go
carolynvs: Huh I wonder what they are doing that it's used? 🤔 

Thanks for checking, I'll try to remember this the next time I'm tempted to remove it!

carolynvs (approved):
Thanks for adding this! I look forward to the day in 2050 when we have all the repos standardized. 😊

PR #140, opened 2023-01-17
Change the API Group from porter.sh/v1 to getporter.org/v1

VinozzZ:
Since we have moved our main domain from porter.sh to getporter.org, we would like to keep things consistent across all projects. Therefore, this PR changes the `apiGroup` from `porter.sh` to `getporter.org`


carolynvs (changes requested):
I found a few files that also need to be updated:

- PROJECT
- tests/integration/suite_test.go (line 134)
- main.go (line 57)

There are also some links that still reference our old v1 site that it would be nice to just quickly fix (release-v1.porter.sh -> getporter.org)

carolynvs (approved):

PR #141, opened 2023-01-24
Allow installing multiple plugins

VinozzZ:
This PR updated the porter dependency and the default porter-agent to v1.0.6. This allows us to take advantage of the new plugin file installing feature.
To match with the new plugins file schema, I had to update the AgentConfig resource definition. This will be a breaking change.


api/v1/agentconfig_types.go
```diff
@@ -207,12 +207,20 @@ func init() {
 	SchemeBuilder.Register(&AgentConfig{}, &AgentConfigList{})
 }
 
+type PluginsSpec struct {
+	// SchemaVersion is the version of the plugins configuration state schema.
+	SchemaVersion string `json:"schemaVersion,omitempty" mapstructure:"schemaVersion,omitempty"`
+
+	// Configs is a map of plugin configuration using plugin name as the key.
+	Configs map[string]Plugin `json:"configs,omitempty" mapstructure:"configs,omitempty"`
```
carolynvs, line 215: I'm a bit confused about why this representation of the document uses `configs` instead of `plugins`?

api/v1/agentconfig_types.go
```diff
@@ -146,10 +150,6 @@ func (ac AgentConfig) MergeReadyConfigs(overrides ...AgentConfig) (AgentConfig,
 	}
 	cfg.Spec = cfgSpec
 
-	if !cfg.Status.Ready {
```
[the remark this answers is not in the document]
VinozzZ: When a AgentConfig has plugins defined, the Ready state should not be checked if the AgentAction is created by AgentConfigController. We already have the check here: https://github.com/getporter/operator/blob/2c4adfab50886f62c70f51c5fe717bd607a3c5e3/controllers/agentaction_controller.go#L517
I added test for this scenario here: https://github.com/getporter/operator/blob/2c4adfab50886f62c70f51c5fe717bd607a3c5e3/controllers/agentaction_controller_test.go#L919

api/v1/agentconfig_types.go
VinozzZ:
```yaml
spec:
  plugins:
    schemaVersion: 1.0.0
    configs:
      kubernetes: {}
```
This is what the schema looks like with the current implementation
If we change from `Configs` to `Plugins`, I find it a bit confusing to have `Plugins` under `Spec.Plugins`

VinozzZ: That's a very good point. I changed it based on your suggestions

api/v1/agentconfig_types.go
```diff
@@ -70,8 +71,11 @@ type AgentConfigSpec struct {
 	// This can be useful for a bundle which is targeting the kubernetes cluster that the operator is installed in.
 	// +optional
 	InstallationServiceAccount string `json:"installationServiceAccount,omitempty" mapstructure:"installationServiceAccount,omitempty"`
+
+	// PluginConfigFile specifies plugins required to run Porter bundles.
+	// In order to utilize mapstructure omitempty tag with an embedded struct, this field needs to be a pointer
 	// +optional
-	Plugins map[string]Plugin `json:"plugins,omitempty" mapstructure:"plugins,omitempty"`
+	PluginConfigFile *PluginFileSpec `json:"pluginConfigFile,omitempty" mapstructure:"pluginsConfigFile,omitempty"`
```
carolynvs, line 78: You are using both `plugins` plural and singular here in the json/mapsturcture tags. I'm not sure which one you want to use?

api/v1/agentconfig_types.go
```diff
@@ -207,6 +207,14 @@ func init() {
 	SchemeBuilder.Register(&AgentConfig{}, &AgentConfigList{})
 }
 
+type PluginFileSpec struct {
+	// SchemaVersion is the version of the plugins configuration state schema.
+	SchemaVersion string `json:"schemaVersion" yaml:"schemaVersion"`
+
+	// Plugins is a map of plugin configuration using plugin name as the key.
+	Plugins map[string]Plugin `json:"configs,omitempty" mapstructure:"configs,omitempty"`
```
carolynvs, line 215: I thought you were going to make it match the suggested schema from my last comment but this wasn't updated to be `plugins` so I'm a bit confused if you intended to match the suggestion or ?

VinozzZ: oops, the rebasing with the latest main really messed it up. I will fix it

carolynvs: Next time, if you leave the rebase for after the review is done, it's easier for me to review since I don't have to go over the entire thing again from scratch.

carolynvs (approved):
Looks great, thanks! 💖

Issue #124, opened 2022-10-10
Update CRDs to use the same naming conventions as in Porter's resources


carolynvs:
Whenever we embed a porter resource inside a CRD, it would help people move between the two (porter cli and the operator) if resource naming conventions were the same between the two. Porter uses lower case words separated by hyphens (e.g. log-level), while the CRDs currently are set up to use camelCase (e.g. logLevel).

The yaml tags are set up so that when working with a resource in json, we use camelCase and when exporting to yaml (to pass a document to Porter) we automatically switch it back to what Porter understands.

I was hoping that we could support "you can copy/paste a Porter resource into the spec of the CRD and it just works". What we have now I think will contribute to hard to spot problems when people forget that we used different naming conventions in the CRD than in Porter.

What do you think about switching the CRDs to use the same naming conventions as Porter?

carolynvs:
Adding @sgettys examples of what it should look like and what's impacted from #125 

ex:
```
apiVersion: porter.sh/v1
kind: PorterConfig
metadata:
  name: porterconfig-sample
  labels:
    porter.sh/testdata: "true"
spec:
  verbosity: debug
  defaultSecretsPlugin: kubernetes.secrets
  defaultStorage: in-cluster-mongodb
  storage:
    - name: in-cluster-mongodb
      plugin: mongodb
      config:
        url: "mongodb://mongodb.porter-operator-system.svc.cluster.local"
```

should become

```
apiVersion: porter.sh/v1
kind: PorterConfig
metadata:
  name: porterconfig-sample
  labels:
    porter.sh/testdata: "true"
spec:
  verbosity: debug
  default-secrets-plugin: kubernetes.secrets
  default-storage: in-cluster-mongodb
  storage:
    - name: in-cluster-mongodb
      plugin: mongodb
      config:
        url: "mongodb://mongodb.porter-operator-system.svc.cluster.local"
```

Update the following values in the spec:
* buildDriver -> build-driver
* defaultStorage -> default-storage
* defaultSecrets -> default-secrets
* defaultStoragePlugin -> default-storage-plugin
* defaultSecretsPlugin -> default-secrets-plugin

This will be a breaking change

PR #142, opened 2023-01-30
change porter config crd to use the same naming conventions as Porter

VinozzZ:
In order for a user to be able to just copy/paste a Porter resource into the spec of the CRD and it just works, the naming for PorterConfig CRD needs to be updated from camel case to seperated by hyphen.
relates to: #124

### DIFF
--- a/config/samples/_v1_porterconfig.yaml
+++ b/config/samples/_v1_porterconfig.yaml
@@ -6,8 +6,8 @@ metadata:
     getporter.org/testdata: "true"
 spec:
   verbosity: debug
-  defaultSecretsPlugin: kubernetes.secrets
-  defaultStorage: in-cluster-mongodb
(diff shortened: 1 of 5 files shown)


config/samples/_v1_porterconfig.yaml
```diff
@@ -6,8 +6,8 @@ metadata:
     getporter.org/testdata: "true"
 spec:
   verbosity: debug
-  defaultSecretsPlugin: kubernetes.secrets
-  defaultStorage: in-cluster-mongodb
+  default-secretsPlugin: kubernetes.secrets
```
carolynvs, line 9: Seeing this sample here has made me realize that we don't have a schemaVersion/Type for config files. I've created https://github.com/getporter/porter/issues/2542 as a placeholder for us to consider that.

carolynvs (approved):
Thanks!

PR #152, opened 2023-02-08
fix race condition to createEmptyPluginVolume

VinozzZ:
# What does this change

Sometimes, a test in our CI build randomly times out when trying to install a plugin. see the log here: https://github.com/getporter/operator/actions/runs/4116258618/jobs/7106066540

I found that it's due to a race condition if two AgentConfig reconcile requests comes in around the same time before a AgentAction is created for installing a plugin.
Currently, this will cause two empty temporary plugin persistent volume claims being created. Then the rename for the plugin pvc will never proceed.

This PR fix the check that determine whether a plugin volume already exists. 

# What issue does it fix

# Notes for the reviewer

# Checklist
- [x] Did you write tests?


carolynvs (approved):
I'm so glad you caught this! 👍

Issue #149, opened 2023-02-02
Make porter agent job retries configurable


carolynvs:
Right now we have job retries disabled while developing but it's a core part of k8s and it should be turned on by default. Let's pick a reasonable default to add to the AgentConfig and then allow people to customize how porter retries the agent action jobs.

VinozzZ:
I tried to enable job retries through using the default `BackoffLimit` on the Job spec as well as setting the `RestartPolicy` to `OnFailure` on the PodSpec created by AgentAction job.

A failed AgentAction job was able to retry by restarting the failed pod up to the back off limit before the AgentAction job is considered as Failed. However, the issue I ran into is that all failed pods are deleted after the backoff limit is reached.

I found a good explanation on this here: https://github.com/kubernetes/kubernetes/issues/74848#issuecomment-971487582
Basically if a Job is marked as failure, the pod has to be deleted when `RestartPolicy` is set to `OnFailure` to prevent the pod keeps restarting. Is this an acceptable default behavior that we want?

carolynvs:
> We could keep the RestartPolicy on the pod as Never and have the AgentAction job doing retry by creating new pod and leave the failed ones alone.

I've had good success with that in other projects I have worked on. 👍

> The default retry limit is 6 on a job, maybe we can set it to 3 to reduce the amount of resource consumption?

Let's keep the default but allow people to change the retry limit on the AgentConfig.

PR #153, opened 2023-02-08
enable retry for agent action job

VinozzZ:
# What does this change

This PR enables retry mechanism on AgentAction jobs by using the default value for `BackoffLimit` from kubernetes.

# What issue does it fix
Closes: #149 

# Notes for the reviewer

# Checklist
- [x] Did you write tests?
- [x] Did you write documentation?
- [x] Did you make any API changes? Update the corresponding API documentation.


controllers/agentaction_controller.go
```diff
@@ -398,6 +398,11 @@ func (r *AgentActionReconciler) createAgentJob(ctx context.Context, log logr.Log
 	labels := r.getAgentJobLabels(action)
 	env, envFrom := r.getAgentEnv(action, agentCfg, pvc)
 	volumes, volumeMounts := r.getAgentVolumes(ctx, log, action, agentCfg, pvc, configSecret, workdirSecret, imgPullSecret)
+	var backoffLimit *int32
+	backoffLimitValue, exist := agentCfg.GetRetryLimit()
+	if exist {
+		backoffLimit = pointer.Int32Ptr(int32(backoffLimitValue))
```
carolynvs, line 404: Can we define RetryLimit as int32 on the AgentConfig so that we don't need to do this casting?

api/v1/agentconfig_types.go
```diff
@@ -311,6 +315,14 @@ func (c AgentConfigSpecAdapter) GetInstallationServiceAccount() string {
 	return c.original.InstallationServiceAccount
 }
 
+// SetRetryAnnotation flags the resource to retry its last operation.
+func (c *AgentConfigSpecAdapter) GetRetryLimit() (int, bool) {
```
carolynvs, line 319: If this function returned a *Int32, we could assign it directly to the JobSpec without as many checks and conditions in the controller. Is using a bool here instead of nil is providing value over returning a single *Int32?

controllers/agentaction_controller.go
```diff
@@ -439,7 +444,7 @@ func (r *AgentActionReconciler) createAgentJob(ctx context.Context, log logr.Log
 						},
 					},
 					Volumes:            volumes,
-					RestartPolicy:      "Never", // TODO: Make the retry policy configurable on the Installation
+					RestartPolicy:      "Never",
```
carolynvs, line 447: Can you include a comment with a link to that great k8's comment you linked in the associated issue for this PR that explains how RestartPolicy=Never works. That was some really good context that will help out whoever works on this code next.

controllers/agentaction_controller_test.go
```diff
@@ -696,7 +696,6 @@ func TestAgentActionReconciler_createAgentJob(t *testing.T) {
 	assertContains(t, job.Labels, porterv1.LabelJobType, porterv1.JobTypeAgent, "incorrect label")
 	assertContains(t, job.Labels, "testLabel", "abc123", "incorrect label")
 	assert.Equal(t, pointer.Int32Ptr(1), job.Spec.Completions, "incorrect job completions")
-	assert.Equal(t, pointer.Int32Ptr(0), job.Spec.BackoffLimit, "incorrect job back off limit")
```
carolynvs, line 699: How about we assert that it is nil now instead of 0?

tests/integration/utils_test.go
```diff
@@ -289,11 +297,29 @@ func getAgentActionCmdOut(action *porterv1.AgentAction, aaOut string) string {
 that does "porter credentials list" will return just the result of the porter command from the job logs. This can be
 used to run porter commands inside the cluster to validate porter state
 */
-func runAgentAction(ctx context.Context, actionName string, namespace string, cmd []string) string {
-	aa := newAgentAction(namespace, actionName, cmd)
+func runAgentActionWithDefaultAgentCfg(ctx context.Context, actionName string, namespace string, cmd []string) string {
+	return runAgentAction(ctx, actionName, namespace, "", cmd)
+}
+
+func runAgentAction(ctx context.Context, actionName string, namespace string, agentCfgName string, cmd []string) string {
```
carolynvs, line 304: I am not seeing where we use this function other than runAgentActionWithDefaultConfig. Did you have plans to support specifying different config and then ended up not needing it?

carolynvs (commented):
I don't see any changes to the controller logic where cascade a job's status to the AgentAction status. Does that mean that when a job fails, and the retryLimit hasn't been reached that the job doesn't register as "completed" in the controller and so we ignore the job's status until it runs out of retries?

It would be great if you could add extra steps in TestAgentActionReconciler_Reconcile (or make a new similar test) to simulate a retry and that the status doesn't change until the retries run out or the job completes successfully.

tests/integration/utils_test.go
VinozzZ: Oh, I'm using it here: https://github.com/getporter/operator/pull/153/files#diff-50115afe52184bcb5d2d1627f4782a72c843b05c99174a6b2a4efe0525a2be66R34

carolynvs: Oops, sorry I missed that. 👍

VinozzZ:
> TestAgentActionReconciler_Reconcile

Yes, the job will only be updated with a Completed/Failed status if the backoff limit has reached. I tried to mimic the scenario in the test you mentioned. Let me know what you think

controllers/agentaction_controller.go
```diff
@@ -401,7 +401,7 @@ func (r *AgentActionReconciler) createAgentJob(ctx context.Context, log logr.Log
 	var backoffLimit *int32
```
carolynvs: I think this can be simplified a lot more to just setting the value on the spec directly using the value from the agent config.

```
porterJob.Spec.BackoffLimit = agentCfg.RetryLimit
```

I don't think we need the GetRetryLimit function, or the checks for existence, etc. If the user set the value to a non-nil value, the job will use the user-define limit. Otherwise nil will be assigned, and the k8s default (6) used instead.

VinozzZ: I have updated the code to not have the checks. However, the `GetRetryLimit` function is needed since the `AgentConfigSpecAdapter` is being used here instead of the raw `AgentConfigSpec`

carolynvs (approved):
I'm excited to see this finally ship! 🎉

PR #156, opened 2023-02-15
155 support multiple architecture image publish

### LINKED ISSUE #155: Support multiple architecture image publish
**Is your feature request related to a problem? Please describe.**
The operator does not run native on MacBook M1's.  When using arm kind to deploy a cluster rosetta2 is not used.

**Describe the solution you'd like**
Build and publish multi-arch versions of images so the platform can pick the appropriate one.

bdegeeter:
# What does this change
* Adding a final multi arch image build for arm and x86
* Added logic to final CI step supporting publish from a fork

CI changes only

# What issue does it fix
Closes #155

# Notes for the reviewer
My goal was not to change any existing test logic.  Some refactoring to the kind test env would
allow a `docker buildx bake` pattern to be used in more places.

# Checklist
- [ ] Did you write tests?
- [x] Did you write documentation?  (added note in gitlab actions with instructions to publish from fork)

### DIFF
--- a/magefiles/magefile.go
+++ b/magefiles/magefile.go
@@ -198,7 +198,7 @@ func getMixins() error {
 
 // Publish the operator and its bundle.
 func Publish() {
-	mg.Deps(PublishImages, PublishBundle)
+	mg.Deps(PublishMultiArchImages, PublishBundle)
 }
 
 // Push the porter-operator bundle to a registry. Defaults to the local test registry.
@@ -367,6 +367,16 @@ func PublishImages() {
 
 	log.Println("Pushing", imgPermalink)
 	must.RunV("docker", "push", imgPermalink)
+}
+
+func PublishMultiArchImages() {
+	meta := releases.LoadMetadata()
+	img := Env.ManagerImagePrefix + meta.Version
+	imgPermalink := Env.ManagerImagePrefix + meta.Permalink
+
+	log.Printf("Multi-arch build and push of %s and %s\n", img, imgPermalink)
+	must.RunV("docker", "buildx", "create", "--use")
+	must.RunV("docker", "buildx", "bake", "-f", "docker-bake.json", "--push", "--set", "porter.tags="+img, "--set", "porter.tags="+imgPermalink, "porter")
 }
 
 func PublishLocalPorterAgent() {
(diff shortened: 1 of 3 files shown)


magefiles/magefile.go
```diff
@@ -369,6 +369,16 @@ func PublishImages() {
...
 }
 
+func PublishMultiArchImages() {
+	meta := releases.LoadMetadata()
+	img := Env.ManagerImagePrefix + meta.Version
+	imgPermalink := Env.ManagerImagePrefix + meta.Permalink
+
+	log.Printf("Multi-arch build and push of %s and %s\n", img, imgPermalink)
+	must.RunV("docker", "buildx", "create", "--use")
+	must.RunV("docker", "buildx", "bake", "-f", "docker-bake.json", "--push", "--set", "porter.tags="+img, "--set", "porter.tags="+imgPermalink, "porter")
```
carolynvs, line 379: What do the porter.tags do? Just wondering who uses the tags and if they get published to the registry or baked into the image?

bdegeeter: Basically it's the equivalent of the `-t` arg to set tags for a build.  This is the beginnings of the value-add of the bulidx bake workflow.  A baker file (docker-bake.json) can contain the definition of docker builds of multiple patterns.  In this case the only target in the file is "porter" which is also the default.  (perhaps it should be changed to `porterop`? )  The [buildx baker](https://docs.docker.com/engine/reference/commandline/buildx_bake/)  pattern has the advantage of managing parallel builds as well.

The baker file can have targets with pre-defined tags, build context, etc.. Anything that would normally be part of a docker build.  The `--set` command is a general purpose arg for CRUD on the baker file at runtime.  Sort of like helm.

In this case there's no tags in the baker file for the `porter` target.  They are being set a runtime.  I see this as a starting place to have a single definition for build and publish patterns that can replace more of the docker build commands over time. I had an iteration of this work using it in more areas, but it would require finer grain mage targets to know when a image build is needed for a local kind load for test runs.  Multi-arch builds cannot load an image into the hosts cache for use by kind.   I wanted to get a final multi-arch publish in as an initial step. 

If this path looks desirable we can look at adopting it in the porter project as well.

carolynvs: Neat! Yeah I'd be down for using this elsewhere. 👍

carolynvs (approved):

Issue #17, opened 2021-03-03
Install the operator into an arbitrary namespace


carolynvs:
Right now the kustomize files generated by the operator sdk hard code the namespace. Let's see if we can ideally make that configurable, so that if we made a helm chart from the kustomize output, the namespace could be set when it's installed. 

It may make sense to generate a helm chart in this repo as well. We can then use helm to install the operator in a bundle, or with a helm chart variant for installing it on a cloud provider.

VinozzZ:
related issue: https://github.com/getporter/operator/issues/75

SubZer0MS:
looking into this one

carolynvs:
Thanks I've assigned the issue to you 👍

SubZer0MS:
Making a note here:

1. Ability to install the operator using kustomize, and not a porter bundle.
2. The destination namespace for the porter operator is hard-coded to "porter-operator-system" and in either case (installing via porter or directly with kustomize) the user should be able to install into their desired namespace.

PR #191, opened 2023-03-30
ability to install Operator with an alternate namcespace than the default (porter-operator-system)

SubZer0MS:
# What does this change
Added functionality of being able to specify an alternate namespace than the default (porter-operator-system) where the Operator is installed by passing in "--param operatorNamespace=operator-ns-name-here" to the operator porter install command.

# What issue does it fix
Closes # [_(issue)_](https://github.com/getporter/operator/issues/17)

# Notes for the reviewer
I'll handle the second issue which is discussed there (ability to install Operator with kustomize, without porter) in a separate PR.

# Checklist
- [X] Did you write tests?
- [X] Did you write documentation?
- [X] Did you make any API changes? Update the corresponding API documentation.

### DIFF
--- a/controllers/agentaction_controller.go
+++ b/controllers/agentaction_controller.go
@@ -483,8 +483,6 @@ func (r *AgentActionReconciler) createAgentJob(ctx context.Context, log logr.Log
 func (r *AgentActionReconciler) resolveAgentConfig(ctx context.Context, log logr.Logger, action *porterv1.AgentAction) (porterv1.AgentConfigSpecAdapter, error) {
 	log.V(Log5Trace).Info("Resolving porter agent configuration")
 
-	operatorNamespace := operatorNamespaceDefault
-
 	logConfig := func(level string, config *porterv1.AgentConfig) {
 		if config == nil || config.Name == "" {
 			return
@@ -495,8 +493,6 @@ func (r *AgentActionReconciler) resolveAgentConfig(ctx context.Context, log logr
 			"namespace", config.Namespace,
 			"name", config.Name,
 			"plugin", config.Spec.PluginConfigFile)
-
-		operatorNamespace = config.Namespace
 	}
 
 	// Read agent configuration defined at the system level
@@ -554,8 +550,6 @@ func (r *AgentActionReconciler) resolveAgentConfig(ctx context.Context, log logr
 func (r *AgentActionReconciler) resolvePorterConfig(ctx context.Context, log logr.Logger, action *porterv1.AgentAction) (porterv1.PorterConfigSpec, error) {
 	log.V(Log5Trace).Info("Resolving porter configuration file")
 
-	operatorNamespace := operatorNamespaceDefault
-
 	logConfig := func(level string, config *porterv1.PorterConfig) {
 		if config == nil || config.Name == "" {
 			return
@@ -564,8 +558,6 @@ func (r *AgentActionReconciler) resolvePorterConfig(ctx context.Context, log log
 			"level", level,
 			"namespace", config.Namespace,
 			"name", config.Name)
-
-		operatorNamespace = config.Namespace
 	}
 
 	// Provide a safe default config in case nothing is defined anywhere
--- a/controllers/agentaction_controller_test.go
+++ b/controllers/agentaction_controller_test.go
@@ -928,7 +928,7 @@ func TestAgentActionReconciler_NoPluginsSpecified(t *testing.T) {
 
 func TestAgentActionReconciler_resolveAgentConfig(t *testing.T) {
 	systemCfg := porterv1.AgentConfig{
-		ObjectMeta: metav1.ObjectMeta{Name: "default", Namespace: operatorNamespaceDefault},
+		ObjectMeta: metav1.ObjectMeta{Name: "default", Namespace: operatorNamespace},
 		Status: porterv1.AgentConfigStatus{
 			Ready: true,
 		},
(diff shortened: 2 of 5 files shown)


carolynvs:
https://getporter.org/contribute/guide/#signing-your-commits explains how to get the DCO check to pass

installer/helpers.sh
carolynvs, line 1: The `configureNamespace` function is used by the configureNamespace custom action for the bundle, it sets up a namespace for use with the operator with required configuration such as a porter configuration file. By default we use the contents of installer/manifests/namespace/defaults/porter-config-spec.yaml. In that file, we reference the in-cluster mongodb service that is installed in the operator namespace and it's hard-coded to porter-operator-system.

The configureNamespace helper function should also update that file when a custom porter configuration file was not specified to use the custom operator namespace.

```yaml
storage:
  - name: "in-cluster-mongodb"
    plugin: "mongodb"
    config:
      url: "mongodb://mongodb.TODO-OPERATOR-NAMESPACE.svc.cluster.local"
```

installer/helpers.sh
```diff
@@ -1,6 +1,23 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+OPNAMESPACE="porter-operator-system"
+
+setCustomNamespaceForOperator() {
+  if [ -z "$1" ]; then
+    echo "No namespace specified, using default $OPNAMESPACE"
```
carolynvs, line 8: Let's always be clear which namespace we are referring to in these messages, because we have both the "porter operator namespace" and also a namespace that is created and configured by the `configureNamespace` custom action in the bundle.

```suggestion
    echo "No porter operator namespace specified, using default $OPNAMESPACE"
```

installer/helpers.sh
```diff
@@ -1,6 +1,23 @@
...
+  else
+    OPNAMESPACE=$1
+    echo "Using custom namespace $OPNAMESPACE"
```
carolynvs, line 11:
```suggestion
    echo "Using custom porter operator namespace $OPNAMESPACE"
```

installer/helpers.sh
```diff
@@ -1,6 +1,23 @@
...
+  fi
+
+  # Replace the namespace in the operator.yaml
+  echo "Setting namespace to $OPNAMESPACE"
```
carolynvs, line 15:
```suggestion
  echo "Setting porter operator namespace to $OPNAMESPACE"
```

installer/vanilla.porter.yaml
```diff
@@ -68,6 +68,13 @@ parameters:
     default: "13.6.2"
     applyTo:
       - install
+  - name: operatorNamespace
+    description: Namespace to install the operator into
+    type: string
+    default: "porter-operator-system"
+    applyTo:
+      - install
+      - upgrade
```
carolynvs, line 77: The `configureNamespace` custom action will also need the porter operator namespace parameter. 
```suggestion
    applyTo:
      - install
      - upgrade
      - configureNamespace
```

installer/vanilla.porter.yaml
```diff
@@ -88,6 +95,12 @@ mixins:
   - kubernetes
 
 install:
+  - exec:
+      description: "Set custom namespace for operator if present"
```
carolynvs, line 99:
```suggestion
      description: "Set custom namespace for operator
```

The parameter `operatorNamespace` will always be populated. So if it's not customized by the user it will just have the default value. Since this step always runs and the helper script always runs kustomize, let's make it clear that this logic always runs.

installer/helpers.sh
```diff
@@ -1,6 +1,23 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+OPNAMESPACE="porter-operator-system"
```
carolynvs, line 4: This works but I wanted you to know about a porter feature in case that affects how you'd like to implement this. 

The parameter operatorNamespace is always populated, it will have the default value if not specified by the user, and a corresponding environment variable, `OPERATORNAMESPACE` set. There isn't a need to define OPNAMESPACE, pass the parameter as an argument to each helper function, and initialize OPNAMESPACE in each function. You can just use OPERATORNAMESPACE.

magefiles/magefile.go
```diff
@@ -334,7 +334,7 @@ func Deploy() {
 		buildPorterCmd("credentials", "apply", "hack/creds.yaml", "-n=operator").Must().RunV()
 	}
 	bundleRef := Env.BundlePrefix + meta.Version
-	installCmd := buildPorterCmd("install", "operator", "-r", bundleRef, "-c=kind", "--force", "-n=operator").Must()
+	installCmd := buildPorterCmd("install", "operator", "-r", bundleRef, "-c=kind", "--force", "-n=operator", "--param", "operatorNamespace="+operatorNamespace).Must()
```
carolynvs, line 337: Let's follow the pattern for allowing the developer to customize development parameters, like PORTER_AGENT_REPOSITORY, using local environment variables instead of requiring changes to the code.

The function `applyHackParameters` should be updated along with the hack/dev-build-params.yaml to set the operatorNamespace parameter from PORTER_OPERATOR_NAMESPACE on the local machine.

carolynvs (changes requested):
Apologies for the slow review, I didn't see that you'd updated the PR after my last comment. It really helps to leave a comment on the PR when you make a change or push a commit that is ready for me to look at as it's easy to miss the notification that a new commit has been pushed to the PR.

I see the tests are not passing, let's get the current batch of comments addressed and see if that resolves it first.

PR #202, opened 2023-04-27
Limit PorterConfig to namespace and system configuration 

### LINKED ISSUE #154: Proposal: Remove resource level config overrides
Remove InstallationSpec.AgentConfig and InstallationSpect.PorterConfig.

**Additional Context**
As we are working on the grpc service to connect outputs from an installation back into kubernetes and allow people to drive events based on when a bundle completes and publishes its outputs, we have realized that letting an installation have config overrides makes it very difficult to select the correct configuration for any one grpc request. Requests always know what namespace they are associated with but do not always have an installation associated to it.

Installation level overrides was a feature I added because it seemed nice, but wasn't driven by a hard requirement or request from anyone. So I'd like to survey our users and see if we can make this change before v1 so that the grpc service can integrate with the operator well.

**Risks/Concerns**
It's not clear yet if anyone relies on this feature.

Adding installation or config references to every grpc call.

**Implementation details**
N/A

# Checklist
- [x] This proposal has remained open for at least one week, to allow time for community feedback.

carolynvs:
# What does this change
Remove the PorterConfig setting from individual resources so that the PorterConfig within any namespace is constant. This allows us to run a porter data service in each namespace and safely work with any resources defined in that namespace.

# What issue does it fix
Fixes #154

# Notes for the reviewer
I was running into build errors because PORTER_OPERATOR_REGISTRY was not set, it really shouldn't be required just to do a build, so I defaulted it to the local registry that mage sets up when it runs the tests, localhost:5000.

# Checklist
- [ ] Did you write tests?
- [x] Did you write documentation?

### DIFF
--- a/mage/env.go
+++ b/mage/env.go
@@ -3,9 +3,6 @@ package mage
 import (
 	"os"
 	"path"
-
-	"github.com/carolynvs/magex/mgx"
-	"github.com/pkg/errors"
 )
 
 const (
@@ -34,7 +31,7 @@ func getAmbientEnvironment() Environment {
 	default:
 		registry := os.Getenv("PORTER_OPERATOR_REGISTRY")
 		if registry == "" {
-			mgx.Must(errors.New("environment variable PORTER_OPERATOR_REGISTRY must be set to push to a custom registry"))
+			registry = "localhost:5000"
 		}
 		return buildEnvironment(name, registry)
 	}
--- a/api/v1/agentaction_types.go
+++ b/api/v1/agentaction_types.go
@@ -11,9 +11,6 @@ type AgentActionSpec struct {
 	// +optional
 	AgentConfig *corev1.LocalObjectReference `json:"agentConfig,omitempty"`
 
-	// PorterConfig is the name of a PorterConfig to use instead of the PorterConfig defined at the namespace or system level.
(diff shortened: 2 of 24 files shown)


sgettys (approved):
LGTM! This will make managing configs much easier. Should there be a docs update to talk about how to manage configs across namespaces?

carolynvs:
@sgettys We have https://getporter.org/operator/administrators/configure-porter-agent/ which I just realized isn't properly linked from the website. I'll add a page like that and then submit a PR to the porter repo to add it to the navigation.

sgettys:
Update the https://github.com/getporter/operator/blob/main/docs/content/glossary.md#porterconfig section of the glossary?

mage/env.go
```diff
@@ -34,7 +31,7 @@ func getAmbientEnvironment() Environment {
 	default:
 		registry := os.Getenv("PORTER_OPERATOR_REGISTRY")
 		if registry == "" {
-			mgx.Must(errors.New("environment variable PORTER_OPERATOR_REGISTRY must be set to push to a custom registry"))
+			registry = "localhost:5000"
```
bdegeeter, line 34: What was the reason for add the default value here?  Does the match the porter magefile pattern?

PR #211, opened 2023-06-09
Create tests for method `renamePluginVolume`

troy0820:
# What does this change
- [x] Add tests for method `renamePluginVolume` 
- [x] Fixes linter nits 

# What issue does it fix
No issue to close but improves the testing posture of this controller. 

# Notes for the reviewer

During adding these tests, there is a condition I believe we are not checking for.  If the list of PVCs returned length is < 1 we still continue to return the `readyPVC, tempPVC, nil` (`r.getExistingPluginPVCs(...)`).  There is a no-op check there for the items list but we are returning:

```go
   return readyPVC, tempPVC, nil
```
even if the list (which doesn't error, because you can return an empty list) results.Items < 1.  Not sure what behavior we want from this but guarding against that was noted during testing this method. If we do want to make sure that we always get a list > 1 from that method, I can update this PR to modify that.  

# Checklist
- [x] Did you write tests?
     - No, it's not needed for adding tests 
     - No, no changes to the API was made

### DIFF
--- a/controllers/agentconfig_controller.go
+++ b/controllers/agentconfig_controller.go
@@ -584,7 +584,7 @@ func (r *AgentConfigReconciler) isReadyToBeDeleted(ctx context.Context, log logr
 func (r *AgentConfigReconciler) getExistingPluginPVCs(ctx context.Context, log logr.Logger, agentCfg *porterv1.AgentConfigAdapter) (readyPVC *corev1.PersistentVolumeClaim, tempPVC *corev1.PersistentVolumeClaim, err error) {
 	results := &corev1.PersistentVolumeClaimList{}
 	err = r.List(ctx, results, client.InNamespace(agentCfg.Namespace), client.MatchingLabels(agentCfg.Spec.Plugins.GetLabels()))
-	if err != nil && !apierrors.IsNotFound(err) {
+	if err != nil && !apierrors.IsNotFound(err) || len(results.Items) < 1 {
 		return nil, nil, err
 	}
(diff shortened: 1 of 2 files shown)


sgettys:
In the scenario you noted yes we can get a list of 0 which will result in the tempPVC and readyPVC both being returned as nil. I wouldn't have any issue with putting a guard check in there as that won't change the behavior of the function at all but it will make it more explicit.

controllers/agentconfig_controller.go
```diff
@@ -588,10 +584,9 @@ func (r *AgentConfigReconciler) isReadyToBeDeleted(ctx context.Context, log logr
 func (r *AgentConfigReconciler) getExistingPluginPVCs(ctx context.Context, log logr.Logger, agentCfg *porterv1.AgentConfigAdapter) (readyPVC *corev1.PersistentVolumeClaim, tempPVC *corev1.PersistentVolumeClaim, err error) {
 	results := &corev1.PersistentVolumeClaimList{}
 	err = r.List(ctx, results, client.InNamespace(agentCfg.Namespace), client.MatchingLabels(agentCfg.Spec.Plugins.GetLabels()))
-	if err != nil && !apierrors.IsNotFound(err) {
+	if err != nil && !apierrors.IsNotFound(err) || len(results.Items) < 1 {
```
troy0820, line 587: If this is < 1, this error could be nil as the List can return without failing.  _(This matches the behavior if there are no PVCs found and fallthrough line 612)_

sgettys (approved):

Issue #75, opened 2022-02-25
Can we add a helm chart to install the operator?


carolynvs:
The manifests that come out of kubebuilder/operator-sdk are all written for kustomize. I'm not sure if we can generate them as-needed (when we change rbac, add resources, etc) instead of regenerating them on every build, and then add in helm templating? Otherwise the helm chart won't have the standard set of features that we expect of helm, since it won't use variables at all.

Right now the operator's bundle uses kustomize to set the manager's image reference dynamically before installing it, so that it uses the manager image distributed with the bundle.

I wonder if we can use kustomize itself to replace values with templates and then make a workable helm chart that way? 🤔

squillace:
Some background to this PR, which is SADLY not a Carolyn pr anymore. :-(

My argument for helm, bundle, and Kustomize supported operator installations _available_ for anyone to use is as follows. 
1. Carolyn was THE person who believed in inclusion, and for her that was a human, not merely a software principle. I believe this PR indicates that she herself, though she didn't use helm to install, wanted any helm user to be able to walk up and install and test out the porter operator before they mastered either Porter or Kustomize or any other specific tool. 
2. Abstractly, Helm is the single largest deployment and packaging tool in the Kubernetes ecosystem. Per the 2022 CNCF survey data (the latest we have -- https://blog.palark.com/cncf-cloud-native-projects-usage-stats-2022/) Helm is between 45% and 60% of a combination of deployment and packaging tools, down from roughly 75% IIRC in 2021.
>![image](https://github.com/getporter/operator/assets/1422718/a136466e-1c5b-4573-8c5d-77ea8ea94787)
>![image (1)](https://github.com/getporter/operator/assets/1422718/2330c8a6-bc7b-43e5-910b-2536453e488f)
It seems clear that Kustomize is becoming more used during this period, likely as a result of being directly integrated into kubectl as a plugin, but helm continues to outstrip it by wide margin, about 2.5x as much usage. My argument is that having a maintained helm chart to install the operator is pure usability and aids the "getting started" operator user. 

None of this, per #217, requires a) the chart or kustomize files to reside in the same repo (though this is a common approach) and b) none of that requires any specific person who doesn't use it to be forced to maintain it. I'll bring this argument to the community discussion today.

PR #217, opened 2023-06-21
generate helm chart and publish to registry.

jhmadhav:
# What does this change
_Give a summary of the change, and how it affects end-users. It's okay to copy/paste your commit messages._

It Introduces generateHelmChart and publishHelmChart commands to mage to automatically generate helm chart for operator from installer manifests, update image references and push the chart to registry. On Publishhelmchart - the operator chart will be generated from kustomize build, crds will be copied into crds directory under charts/operator, and rest of the manifests into templates folder in charts/operator/templates and Porter operator namespace configuration onboarding yaml's will be copied into templates and proper templating will be applied supporting helm values making a installable helm chart- programmatically. Helm chart version will be updated based on meta.version and operator helm chart will be pushed to registry for consumption. This opens doors to install operator through helm rather than from bundle and since the crds are generated from operator kustomization, manifests need not be maintained for helm explicitly.

on publishHelmChart (from directory perspective on how files are generated)

image

From CLI perspective - on how chart is published (This will be run on mage publish as well)

image

# What issue does it fix
Closes # _(issue)_

_If there is not an existing issue, please make sure we have context on why this change is needed. See our Contributing Guide for [examples of when an existing issue isn't necessary][1]._

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,9 @@ installer/porter.yaml
 *.swo
 *~
 coverage/coverage.html
+
+# Helm
+*.tgz
+!charts/operator
+charts/operator/templates/*.yaml
+charts/operator/crds/*
--- a/.gitignore
+++ b/.gitignore
@@ -35,7 +35,7 @@ installer/porter.yaml
 *~
 coverage/coverage.html
 
-# Helm
+# Helm ignore
 *.tgz
 !charts/operator
 charts/operator/templates/*.yaml
--- a/charts/operator/Chart.lock
+++ b/charts/operator/Chart.lock
@@ -1,0 +1,6 @@
+dependencies:
+- name: mongodb
+  repository: https://charts.bitnami.com/bitnami
+  version: 13.6.2
+digest: sha256:1ac9a80f5e4db4e4d98f5cde28ef3bbbb869efb3c045450f79b72bfe620b45f1
+generated: "2023-06-20T15:03:54.087266138-07:00"
--- a/charts/operator/Chart.yaml
+++ b/charts/operator/Chart.yaml
@@ -1,0 +1,13 @@
+apiVersion: v2
+name: porter-operator
+version: v0.0.0
+appVersion: v0.0.0
+description: Deploy components and dependencies of Porter Operator
+home: https://porter.sh
+sources:
+  - https://github.com/getporter/operator
+dependencies:
+  - name: mongodb
+    version: 13.6.2
+    repository: https://charts.bitnami.com/bitnami
+    condition: mongodb.enabled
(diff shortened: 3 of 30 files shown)


jhmadhav:
@bdegeeter updated can you please check

bdegeeter:
I would expect the to be driven from an issue.  There's no issue associated with this PR.  I appreciate your effort and contribution; however, we need to determine how best to host a helm chart for the operator. Aspects such as maintainability and publishing need to be considered as part of a change like this.   A helm chart will need to be versioned and published separately from the operator.  It's not uncommon to see them maintained in a separate versioned git repo. 

We need to consider the need to publish a new version of a helm chart when there no functional change to the operator.  It appears this change is coupling the helm chart version to the porter operator version.  There are patterns for maintaining and hosting helm and kustomize form a single source of true that I’d like to see us review. 

https://www.linkedin.com/pulse/using-helm-chart-kustomize-managing-multiple-microservices/

We want to meet users where they are at when installing the operator which means having a development workflow to produce a helm chart, kustomize and porter bundle for operator installation.  Let's review this and take it from there.

squillace:
Your point is correct, @bdegeeter, and here's the issue: https://github.com/getporter/operator/issues/75. I note that Carolyn was interested a year ago in both helm and kustomize solutions so that users could install things _using the tool that they wanted to use BEFORE they had porter_. That's the only motivation here: to support the widest class of possible users. Carolyn was always into inclusion. :-|

That said, I'll pick up the thread in that issue and give some user data that describes why helm is one of the most important operation tools in the entire ecosystem (which of course is the only reason to support it). 

but NOTE: that is different than merging this PR as-is. I've had a look and I don't actually like the way it's done, so my suggestion for what it's worth is to sit on this, extract the proper method of doing things, and place that implementation somewhere where it doesn't affect other users. There's no hurry here. My two cents......

jhmadhav:
I appreciate your feedback on this. @bdegeeter, @squillace and @schristoff.  I agree with aspects of maintainability as everyone rightly pointed. Thanks @squillace for finding the issue :)  I remember having some discussions related to helm charts with Carlolyn but could not find the issue to link to this PR. I agree that this approach is not viable for generation of complex helm templating, and we need a maintainable way to produce helm charts.  Consider this is just an attempt to quickly generate charts from kustomize without maintaining two sources and apply helm templating to generated manifests, just as described in [Issue: #75](https://github.com/getporter/operator/issues/75). I really appreciate spending some time in community meeting on maintaining helm charts too. @schristoff. I embrace both helm and kustomize, and kustomization is already provided by kubebuilder out of box. Adding to @squillace, supporting helm charts would greatly help users relying on helm way of deploying and would encourage them to try Porter.

schristoff:
I have moved this to [deployments](github.com/getporter/deployments)

PR #219, opened 2023-07-05
chore: Update dependencies client-go and controller-runtime

troy0820:
# What does this change
This updates **client-go to 0.28.3-alpha** while also having to update **controller-runtime to 0.15.0**
When updating it produced the following error: 

```bash
../../go/pkg/mod/k8s.io/client-go@v0.27.3/applyconfigurations/meta/v1/unstructured.go:64:38: cannot use doc (variable of type *"github.com/google/gnostic/openapiv2".Document) as *"github.com/google/gnostic-models/openapiv2".Document value in argument to proto.NewOpenAPIData
```

Both releases here: 
[kubernetes issue with why we need client-go 0.28.0-alpha](https://github.com/kubernetes/client-go/issues/1269)
[controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0)

# What issue does it fix
Closes # [Dependabot PR](https://github.com/getporter/operator/pull/215)
Dependabot PR that tried to do this automatically went to client-go 0.27.0 but we have to skip that to get this to work with the above issues.  

# Notes for the reviewer
Because of the breaking change of controller-runtime, there are some adjustments added to the tests to get them to pass.  
- [x] We have to add `WithStatusSubResource(...)` to the setup functions in the test or else we can't update the status on the reconcile
- [x] We have to remove some assertions that will not persist the objects (`pvc, pv`) as the object tracker can create them in the fake client but not list/get them during operations.  
- [x]  None of the logic has been changed to the controller responsible 
- [x] I'm creating an issue to create smaller tests to coincide with how we expect this to work.  
    - [x] https://github.com/getporter/operator/issues/220 

# Checklist
- [ ] Did you write tests?
     - [x]  No, modified test(s) to work with the uplifted dependency. 
     - [x]  No, documentation is not necessary 
     - [x] No changes to the api


schristoff (approved):
Will ya throw a link to the issue to make codecov happy _eventually_

troy0820:
https://github.com/getporter/operator/issues/220 <-- Issue link.  
https://github.com/getporter/operator/pull/221 <-- Issue PR (increases the coverage but not the lifecycle of the controller, so it partially resolves this issue) 
@schristoff

PR #245, opened 2023-08-09
(feat) PR for Installation outputs GRPC Service 

### LINKED ISSUE #63: Expose bundle outputs inside a cluster
The outputs generated from running an installation should be natively accessible from Kubernetes. So after the installation completes the output values should be copied into a secret.

We have a couple options for output syncing:

1. Create a new secret for the outputs after each run. This more faithfully represents how porter works (outputs are immutable and tied to a porter run), but puts more decisions on the consumer. Such as do they want the most recent value of an output, or do they want the output generated by a specific generation of the installation.
2. Keep a single secret per installation and update it with the outputs after each run. This would be easier for people to consume but perhaps oversimplifies things.

I have added https://github.com/getporter/porter/issues/1877 to porter so that we can correlate an installation spec generation with a particular run (and its outputs).

troy0820:
# What does this change
- [x] ~~Create the installationoutputs custom resource definition via api/v1~~
- [x] ~~Create the intallationoutoutputs  deep copy functions~~
- [x] ~~Add additional printer columns~~
- [x] ~~Add condition on installationoutputs custom resource~~
- [x] ~~Add the necessary rbac for installation controller to mess with installationoutputs~~
- [x] ~~Add the roles (generated) with the installationoutputs~~

### Above is now merged but leaving for visibility

- [x]  Add the porter grpc client to the installation controller struct (to query the grpc endpoint)
- [x] Add owner reference and options in for `Owns(&v1.InstallationOutput{}, builder.MatchEveryOwner)`
- [x] Hardcoded service to client connection in main.go (TODO: fix this or make sure we don't do this if we can't establish connection) 
- [x] Do not fail if we can't get the outputs from the grpc server 
- [x] Requeue if outputs is deleted and recreate
- [x] Add annotation on installation resource that outputs were made/created `getporter.org/installationoutput: "true"`
- [ ] Add porter installation for operator to launch service when the operator installs. 
   - [ ] This is moving to getporter/deployment but also being made a helm chart.  
- [ ] Add test to demonstrate flow 
- [ ] ~~Make sure outputs get updated when install updates~~ _(deferred until after v1.0.0)_
- [x] Clean up code _(get rid of TODOs)_
- [ ] ~~Waiting on a bundle for upgrade to test logic~~

# What issue does it fix
Closes #63 

# Notes for the reviewer
What is struck through is being handled by PR https://github.com/getporter/operator/pull/246 to lessen the load when reviewing.
```python
NAME                      PORTER NAME               PORTER NAMESPACE   PHASE
test-install-1691504789   test-install-1691504789   demo               Succeeded
```

```python
NAMESPACE   NAME                      PORTER NAME               PORTER NAMESPACE   PHASE       OUTPUT NAMES
demo        test-install-1691504789   test-install-1691504789   demo               Succeeded   outAction,outDelay,outExitStatus,outInsecureValue
```
```python
apiVersion: v1
items:
- apiVersion: getporter.org/v1
  kind: InstallationOutput
  metadata:
    creationTimestamp: "2023-08-09T18:55:48Z"
    generation: 1
    name: test-install-1691504789
    namespace: demo
    ownerReferences:
    - apiVersion: getporter.org/v1
      kind: Installation
      name: test-install-1691504789-nf9k2
      uid: a7d5ba66-c4e9-4a3c-9241-d6957e46aa72
    resourceVersion: "1119303"
    uid: aee802c4-01e8-418b-8266-628665807fbd
  spec:
    name: test-install-1691504789
    namespace: demo
  status:
    conditions:
    - lastTransitionTime: "2023-08-09T18:55:48Z"
      message: outputs custom resource generated succeeded
      reason: InstallationOutputCreatedSuccess
      status: "True"
      type: InstallationOutputSucceeded
    outputs:
    - name: outAction
      sensitive: false
      type: string
      value: install
    - name: outDelay
      sensitive: false
      type: integer
      value: "3"
    - name: outExitStatus
      sensitive: false
      type: integer
      value: "0"
    - name: outInsecureValue
      sensitive: false
      type: string
      value: default
    phase: Succeeded
```
# Checklist
- [x] Did you write tests?
- [x] Did you make any API changes? Update the corresponding API documentation.
     - [x] Yes, added a new type to `api/v1` called `InstallationOuput`.  No api documentation change is necessary.


controllers/installation_controller.go
```diff
@@ -143,13 +156,116 @@ func (r *InstallationReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	}
+	install.Status.Outputs = outputs
+	install.Status.OutputNames = strings.Join(outputNames, ",")
+
+	return install, nil
+}
+
+func (r *InstallationReconciler) CreateInstallationOutputsCR(ctx context.Context, install *v1.Installation, in *installationv1.ListInstallationLatestOutputResponse) (*v1.InstallationOutput, error) {
+	if len(in.Outputs) < 1 {
+		return nil, fmt.Errorf("no outputs for the installation %s", install.Name)
```
schristoff, line 250: Instead of error, maybe best to log?
Some installations may not have outputs

schristoff: Could the installationoutput cr have an owner ref to the installation?

main.go
```diff
@@ -62,11 +66,19 @@ func main() {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
-
+	// TODO: Set this with credentials on the server side
+	// TODO: This needs to be set so we don't fail when instantiating client
```
schristoff, line 70: I feel like this is most important todo, the other ones we can live with

schristoff:
I'm moving this out of draft :)

controllers/installation_controller.go
troy0820: > Instead of error, maybe best to log?

This only makes the CR if the outputs exist from the previous call.  

> Some installations may not have outputs

It errors only to requeue in the event that the output length being returned from the grpc server is less than 1 .  I agree, I should probably still log.  I have to short circuit in the event that the output doesn't exist or it'll make a CR with a blank output.

main.go
troy0820: Yeah, I was thinking tls and generate key/pair before creating the server, however that is done outside of this PR to manipulate the grpc server.  _(The GRPC service is a deployment outside of the controller-manager)_.  The client setting up however, I can pass it credentials to interact with this.  I can clean up the TODO but need more cycles to do the grpc server credentialing on the server side.

main.go
```diff
@@ -62,11 +66,19 @@ func main() {
...
+	conn, err := grpc.DialContext(context.Background(), "porter-grpc-service:3001", grpc.WithTransportCredentials(insecure.NewCredentials()))
```
troy0820, line 71: What I can do to not be dependent on the gRPC server being there is to only set this if the service is present.  

Basically, don't error if this fails but pass an empty listener and log "no grpc client for installationoutputs" @schristoff

schristoff: Sgtm

main.go
schristoff: Sounds good, we should separate this into it's own issue perhaps

schristoff (approved):
Looking through I see there's already an ownerref for instllation outputs in the `installation_controller.go` - so we should be good to go.

PR #246, opened 2023-08-09
Installationoutputs custom resource

### LINKED ISSUE #63: Expose bundle outputs inside a cluster

troy0820:
# What does this change

Sets up work to be accomplished in https://github.com/getporter/operator/pull/245 which will fully resolve #63. 
This sets up the crd and the rbac associated with the new custom resource `installation outputs`

# What issue does it fix
Partially Closes #63 

# Notes for the reviewer
**NOTE: Most of the work is being done in the draft PR from above.  The PR was getting to big and decided to break it up.**

# Checklist
- [ ] Did you write tests?


api/v1/installationoutput_types.go
```diff
@@ -0,0 +1,61 @@
...
+
+	Outputs []Output `json:"outputs,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+
+// +kubebuilder:printcolumn:name="Porter Name",type="string",JSONPath=".spec.name"
+// +kubebuilder:printcolumn:name="Porter Namespace",type="string",JSONPath=".spec.namespace"
+// +kubebuilder:printcolumn:name="Phase",type="string",JSONPath=".status.phase"
```
devigned, line 40: Would you want to show the outputs with `-o wide`? Might be nice to add the outputs field with a `priority=1`.

devigned (commented):
I had one suggestion, but otherwise lgtm!

api/v1/installationoutput_types.go
troy0820: I don't believe that I can do this being that the column that would display the outputs would need to be able to handle the "array/slice" of "output".  Currently the only supported fields are 
```
integer – non-floating-point numbers
number – floating point numbers
string – strings
boolean – true or false
date – rendered differentially as time since this timestamp
```
I was thinking a list of strings comma separated would do but wouldn't be able to be surfaced from the kubebuilder directive. 

Would the first output be sufficient?

devigned: Yeah, I think you'd have to define another field where you format the string you'd want from the outputs, reconcile it each time, and have the print column reference the new field.

I don't think this should hold up this PR. I was just thinking about how it would feel using kubectl.

troy0820: If it's worth having for the user experience, the handling of that could be valuable for the user. That would be great to add in the WIP PR.  I'm willing to add it there if it makes sense. 

````go
 type InstallationOutputStatus struct {
     Phase AgentPhase `json:"phase,omitempty"`
  
     Conditions []metav1.Condition `json:"conditions,omitempty"`
  
     Outputs []Output `json:"outputs,omitempty"`
     
     OutputNames string `json:"outputNames, omitempty"` // <-- new field with priority=1 for wide output.
  }
}


//logic in reconciler 
//Something like this

names := []string{}

for _, name := range inst.Status.Outputs {
    names = append(names, name.Name)
}
outputNames := strings.Join(names, ",")  

// Add to install.Status, etc 
```

troy0820: Added 
- [x] outputNames as a new field, 
- [x] scaffold CRD 
- [x] added the printercolumn to be priority 1.  

`kubectl get installationoutput -A `

```python
NAMESPACE   NAME                        PORTER NAME   PORTER NAMESPACE   PHASE
default     installationoutput-sample
```

`kubectl get installationoutput -A -o wide`

```python
NAMESPACE   NAME                        PORTER NAME   PORTER NAMESPACE   PHASE   OUTPUT NAMES
default     installationoutput-sample
```

troy0820: More realistic version of what is to be expected:
```python
NAMESPACE   NAME                      PORTER NAME               PORTER NAMESPACE   PHASE       OUTPUT NAMES
demo        test-install-1691504789   test-install-1691504789   demo               Succeeded   outAction,outDelay,outExitStatus,outInsecureValue
```

devigned (approved):
lgtm!

schristoff (approved):

PR #258, opened 2023-08-31
Use SetControllerReference from controllerutil

troy0820:
_This is for testing purposes_


schristoff:
This is part of #248  that will close #54

schristoff (approved):
🎉 Happy to see this

PR #267, opened 2023-09-26
(feat): Fix up operator bundle 

schristoff:
To do:
- [x]  Update Quickstart & documentation to let people test out this upgrade path
- [ ] Build and publish this bundle


installer/vanilla.porter.yaml
```diff
@@ -1,11 +1,19 @@
...
 description: "The Porter Operator for Kubernetes. Execute bundles on a Kubernetes cluster."
 registry: ghcr.io/getporter
 dockerfile: Dockerfile.tmpl
 
 parameters:
+  - name: repository
+    description: Porter Operator Manager Image Repository
+    type: string
+    default: ""
+  - name: digest
```
schristoff, line 13: Is having people pass in their own digest weird? Should we do a tag instead? 🤔

schristoff:
I feel like we should put some super basic steps on getting started w/ the operator in our readme, maybe 🤔

PR #279, opened 2023-10-20
Allow multiplatform build to pull image with arch manifest

troy0820:
# What does this change
Without this, trying to run a demo the image fails because it doesn't have the GOOS/GOARCH in the manifest with k8s.  

# What issue does it fix
_Didn't create one_ 

# Notes for the reviewer

# Checklist
- [ ] Did you write tests?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1.4
 
 # Build the manager binary
-FROM golang:1.20.5 as builder
+FROM --platform=$BUILDPLATFORM golang:1.21.3 as builder
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
@@ -17,9 +17,11 @@ COPY main.go main.go
 COPY --link api/ api/
 COPY --link controllers/ controllers/
 
+ARG TARGETARCH
+ARG TARGETOS
 # Build
 RUN --mount=type=cache,target=/go/pkg/mod --mount=type=cache,target=/root/.cache/go-build \
-    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o manager main.go
+    CGO_ENABLED=0 GOOS=$TARGETOS  GOARCH=$TARGETARCH GO111MODULE=on go build -o manager main.go
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details


troy0820:
docker manifest inspect ghcr.io/getporer/porter-operator:v1.0.0 
```yaml
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.oci.image.index.v1+json",
   "manifests": [
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 2372,
         "digest": "sha256:c05bc591cd0288a9805ebd9f89520013243775a5d9fa691a017dd218ba00fb77",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 2372,
         "digest": "sha256:cc3c904b67da278128481304169268c15a075c83df467adcfdadf64f39e93490",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 566,
         "digest": "sha256:6ff4de306a9101d610f67531191ca32a81021ee114c99ebf39bae6eb0d60dca6",
         "platform": {
            "architecture": "unknown",
            "os": "unknown"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 566,
         "digest": "sha256:79915147b7f5e4daec3546127aaf0d488cb375a1f0fb8ac69f78c11af94898c3",
         "platform": {
            "architecture": "unknown",
            "os": "unknown"
         }
      }
   ]
}
```

Dockerfile
```diff
@@ -17,9 +17,11 @@ COPY main.go main.go
 COPY --link api/ api/
 COPY --link controllers/ controllers/
 
+ARG TARGETARCH
+ARG TARGETOS
 # Build
 RUN --mount=type=cache,target=/go/pkg/mod --mount=type=cache,target=/root/.cache/go-build \
-    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o manager main.go
+    CGO_ENABLED=0 GOOS=$TARGETOS  GOARCH=$TARGETARCH GO111MODULE=on go build -o manager main.go
```
schristoff, line 24: are there any targetOS we dont support? did u run this and double check it on your windows pc?

troy0820: I do believe that we don't support windows, but I did this in wsl2.  This is the same thing as the docker-bake.json that we currently have.  It just does this in the Dockerfile.  I needed this to push up the image but I did that already.  I can remove this and only update the image to 1.21.3.  _(Because we do this in a pipeline, it may be better to not do this if windows isn't supported)_

So when building it, I did this: 
`docker buildx build --push --platform linux/arm64/v8,linux/amd64 -t ghcr.io/getporter/porter-operator:v1.0.0 .` so that could get us the image we needed at v1.0.0 which was having issues because the "platform" was missing on the docker manifest.

schristoff: sgtm, we should note what OSes we support here maybe

schristoff: i dont have a strong opinion on this - im just trying to protect against future confusion

troy0820: Agree, I can remove this and allow the image to source it's contents from what is supported.  I initially only needed this to push the operator image to ghcr.  I have done this and can remove this.

PR #312, opened 2024-03-30
Update Go version and dependencies to ease development

kichristensen:
# What does this change
Because of old Go version and different dependencies not supporting newer Go versions, it takes some work to get things to compile. You will get weird errors when using newer versions of Go, none of them describes what the real issue is.
Updating the Go version to 1.21 allows it to be compiled with newer versions of the Go compiler, but requires different dependencies to be updated.
The Porter version and mixins referenced is also old, so it was updated to the newest versions.
External dependencies of, e.g., Kustomize, controller-gen and Ginkgo was also very old and no longer supported, so they were also updated.

Updates the following Go packages:
* Ginkgo
* Gomega
* Porter
* Porter Magefiles

Mixins:
* Kubernetes
* Helm3
* Exec

External dependencies:
* Operator SDK
* Kustomize
* controller-gen
* Ginkgo

# Checklist
- [ ] Did you write tests?


go.mod
```diff
@@ -1,6 +1,6 @@
 module get.porter.sh/operator
 
-go 1.20
+go 1.21
```
troy0820, line 3: Because you are updating this, it’ll help if you also update controller-runtime to v0.17. This dep is also at go version 1.21

troy0820:
This also means that we have to cut a new version of the operator. Older go versions will not be able to use this version as the new minimum required go version will be 1.21.

bdegeeter:
Are specific features of the newer go version needed?

kichristensen:
> Are specific features of the newer go version needed?

@bdegeeter From my point for view no, this PR is only meant to solve a lot of problems I had when trying to build the project. I had ~3 errors that took a long time to figure out, just because I had Go 1.22 installed. The errors did not mention in anyway that it had something to do with the version of Go.
But @troy0820 might have better insights to you question

bdegeeter:
> @bdegeeter From my point for view no, this PR is only meant to solve a few problems I had when trying to build the project. I had ~3 errors that took a long time to figure out, just because I had Go 1.22 installed. I can remember 2 of them off the top of my head, one was with `staticcheck` during linting and the other was with `controller-gen`.
> The errors did not mention in anyway that it had something to do with the version of Go.

Thanks for clarifying.  I'm running into similar issues with my go 1.22.1 env. IIRC, the version checking of some of the tools for the build could use a bit a refinement.  It would be great to capture some of the specifics so we can make sure to address the root cause. 

Any chance you're running Apple silicon?  Are you getting clean local test runs?  `mage -v Test`?

troy0820 (commented):
Everything seems to compile, but you have a failing test case.

kichristensen:
> > @bdegeeter From my point for view no, this PR is only meant to solve a few problems I had when trying to build the project. I had ~3 errors that took a long time to figure out, just because I had Go 1.22 installed. I can remember 2 of them off the top of my head, one was with `staticcheck` during linting and the other was with `controller-gen`.
> > The errors did not mention in anyway that it had something to do with the version of Go.
> 
> Thanks for clarifying. I'm running into similar issues with my go 1.22.1 env. IIRC, the version checking of some of the tools for the build could use a bit a refinement. It would be great to capture some of the specifics so we can make sure to address the root cause.
> 
> Any chance you're running Apple silicon? Are you getting clean local test runs? `mage -v Test`?

Unfortunately I'm not running on Applie silicon. I was getting clean test runs before upgrading `controller-runtime`. I'm currently looking into the failing test

kichristensen:
> Everything seems to compile, but you have a failing test case.

I'm looking into the failing test now. I found the root cause to be the breaking change, https://github.com/kubernetes-sigs/controller-runtime/pull/2633 introduced in controller-runtime v0.17.0. But still haven't figured out how to fix it, so any good ideas are more than welcome, but I will continue to looking into the issue

kichristensen:
After some more investigation I think I was wrong in my original assumption that it was https://github.com/kubernetes-sigs/controller-runtime/pull/2633 that triggered the failing test instead I think it is https://github.com/kubernetes-sigs/controller-runtime/pull/2484.
Still investigating how to fix it

kichristensen:
All tests are should now be passing

PR #326, opened 2024-05-21
Add deletion policy annotation on all installations 

### LINKED ISSUE #297: Deleting installations should uninstall and keep the Kubernetes behavior consistent with resource garbage collection

Currently when you are deleting an installation resource, the custom resource will be deleted and the installation will be uninstalled.  The behavior from before would uninstall the installation but leave around the custom resource.  In our discussions it was said that we want to keep this behavior consistent with how Kubernetes does resources.  The allowance of deleting a namespace where it would delete all the resources (uninstall and remove custom resources) should not impede the operations that porter operator needs to do to clean up the installations present on the cluster. 

**Additional Context**

Propose that when deleting resources, we remove the installation and its components but allow the state to be saved in the backend _e.g. mongodb, etc_ for reinstallation, etc.  The operator right now lives in a namespace where that can happen but also can be deleted to ruin this implemenation. 

**Risks/Concerns**

1. The risk is that the grpc server outputs will reflect the installation as maybe installed and/or not reflect the current state of the cluster. 
1. Kubernetes doesn't like the way we are doing this and demands we follow the constraints of the cluster 

 
1. Not removing the resource, however that will lead to a situation where the finalizer will never be removed and not allow the namespace to be deleted.  
1. Removing the installation and uninstalling but not saving the state of the installation in mongo.  
 
**Implementation details**

TBD - _Gathering notes from the community meeting and discussing with the maintainers_ 

# Checklist
- [ ] This proposal has remained open for at least one week, to allow time for community feedback.

troy0820:
# What does this change
This will add the annotation that we will use to determine how we **want** to delete the resources (installations and otherwise) to adopt the pattern from cross-plane as noted in issue #297 

_~~This is still in draft~~_

All this does is just add the annotation, the WIP will use this to implement the logic behind this new annotation.

# What issue does it fix
Partially closes #297 

# Notes for the reviewer

# Checklist
- [x] Did you write tests?
- [ ] Did you write documentation? _maybe?_ public facing for sure.
- [ ] Did you make any API changes? Update the corresponding API documentation. _none_

### DIFF
--- a/controllers/installation_controller.go
+++ b/controllers/installation_controller.go
@@ -158,6 +158,13 @@ func (r *InstallationReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	log.V(Log4Debug).Info("Reconciliation complete: A porter agent has been dispatched to apply changes to the installation.")
+	// NOTE:  If this is nil, it will use the default policy of Delete
+	err = r.applyDeletionPolicy(ctx, log, inst, inst.GetAnnotations()[v1.PorterDeletePolicyAnnotation])
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	log.V(Log4Debug).Info("Reconciliation complete: A porter agent has been dispatched to apply changes to the installation.")
 	if r.PorterGRPCClient != nil {
 		return r.CheckOrCreateInstallationOutputsCR(ctx, log, inst)
 	}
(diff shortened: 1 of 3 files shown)


controllers/installation_controller.go
```diff
@@ -404,3 +411,27 @@ func (r *InstallationReconciler) retry(ctx context.Context, log logr.Logger, ins
...
+	}
+
+	if _, ok := annotations[v1.PorterDeletePolicyAnnotation]; !ok {
+		annotations[v1.PorterDeletePolicyAnnotation] = v1.PorterDeletePolicyDelete
+		inst.SetAnnotations(annotations)
+		return r.Update(ctx, inst)
+	}
+
+	if policy != v1.PorterDeletePolicyOrphan && policy != v1.PorterDeletePolicyDelete {
+		annotations[v1.PorterDeletePolicyAnnotation] = v1.PorterDeletePolicyDelete
```
sgettys, line 429: Can you put a trace log in for this case so we can see if we're ever trying to set the delete policy to something we don't expect?

troy0820: Definitely,  I passed in the logger but I never used it 🤕

troy0820: The force push doesn't show that this is updated but it is.  Using a debug log to ensure that the policy they entered is not that of the two policies we have.  Resolving this.

sgettys (approved):

PR #328, opened 2024-06-03
Create grpc connection only on invocation of the CRUD events for the installation controller

troy0820:
Before this change we were creating a persistent connection with the grpc server when starting the controller manager.  This led to problems decoupling that connection when starting the operator.  Now this connection will only  be established if the creation/update of an installation is done.  This will reduce the overhead of that established connection as well as not being able to reestablish this connection on subsequent calls if not done at startup. 

# What does this change

- [x] Creates the grpc connection when the controller is invoked on a CRUD event for the installation controller 
# What issue does it fix

No issue but discussed during the maintainers call 

# Notes for the reviewer
This doesn't impact anything as this image shows that this still works, just changes the implementation around how we establish a connection with the grpc server 
image

# Checklist
- [x] Did you write tests? _~~Existing tests verify nothing has changed.~~_


controllers/installation_controller.go
```diff
@@ -436,3 +441,15 @@ func (r *InstallationReconciler) applyDeletionPolicy(ctx context.Context, log lo
 	inst.SetAnnotations(annotations)
 	return r.Update(ctx, inst)
 }
+
+func createPorterGRPCClient() (porterv1alpha1.PorterClient, error) {
+	conn, err := grpc.DialContext(context.Background(), "porter-grpc-service:3001", grpc.WithTransportCredentials(insecure.NewCredentials()))
```
troy0820, line 446:
```suggestion
	conn, err := grpc.DialContext(ctx, "porter-grpc-service:3001", grpc.WithTransportCredentials(insecure.NewCredentials()))
```

controllers/installation_controller.go
```diff
@@ -165,10 +169,11 @@ func (r *InstallationReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
 	log.V(Log4Debug).Info("Reconciliation complete: A porter agent has been dispatched to apply changes to the installation.")
-	if r.PorterGRPCClient != nil {
-		return r.CheckOrCreateInstallationOutputsCR(ctx, log, inst)
+	r.PorterGRPCClient, err = createPorterGRPCClient()
+	if err != nil {
+		return ctrl.Result{}, nil
 	}
-	return ctrl.Result{}, nil
+	return r.CheckOrCreateInstallationOutputsCR(ctx, log, inst)
```
troy0820, line 176: Probably could add some logging here to identify that this is being done.

controllers/installation_controller.go
```diff
@@ -121,10 +124,11 @@ func (r *InstallationReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 		// Nothing for us to do at this point
 		log.V(Log4Debug).Info("Reconciliation complete: A porter agent has already been dispatched.")
-		if r.PorterGRPCClient != nil {
-			return r.CheckOrCreateInstallationOutputsCR(ctx, log, inst)
+		r.PorterGRPCClient, err = createPorterGRPCClient()
+		if err != nil {
+			return ctrl.Result{}, nil
 		}
-		return ctrl.Result{}, nil
+		return r.CheckOrCreateInstallationOutputsCR(ctx, log, inst)
```
troy0820, line 131: Log here as well,

controllers/installation_controller.go
```diff
@@ -121,10 +124,11 @@ func (r *InstallationReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 		// Nothing for us to do at this point
 		log.V(Log4Debug).Info("Reconciliation complete: A porter agent has already been dispatched.")
-		if r.PorterGRPCClient != nil {
-			return r.CheckOrCreateInstallationOutputsCR(ctx, log, inst)
+		r.PorterGRPCClient, err = createPorterGRPCClient()
+		if err != nil {
```
troy0820, line 128: Probably could debug log to show that we are not failing but not doing outputs because the grpc server is not available.

controllers/installation_controller.go
```diff
@@ -436,3 +445,15 @@ func (r *InstallationReconciler) applyDeletionPolicy(ctx context.Context, log lo
 	inst.SetAnnotations(annotations)
 	return r.Update(ctx, inst)
 }
+
+func createPorterGRPCClient(ctx context.Context) (porterv1alpha1.PorterClient, error) {
+	conn, err := grpc.DialContext(ctx, "porter-grpc-service:3001", grpc.WithTransportCredentials(insecure.NewCredentials()))
```
troy0820, line 450: I don't like that this is hardcoded and trying to resolve what this would be beyond the dns entry in the namespace it lives in.  We can deal with this for now, but this should be configurable.

controllers/installation_controller.go
```diff
@@ -436,3 +445,15 @@ func (r *InstallationReconciler) applyDeletionPolicy(ctx context.Context, log lo
...
+	if err != nil {
+		return nil, fmt.Errorf("error setting up listener for porter grpc client")
+	}
+	defer conn.Close()
```
troy0820, line 454: This is going to close once the function is over, we need to move this defer to the method return and close from there.

controllers/installation_controller.go
troy0820: Maybe get the service of the porter-grpc-service (deployment should name this) and then use the service name/port? I don't want to make the assumption that this is what we are naming this.

controllers/installation_controller.go
```diff
@@ -165,10 +173,15 @@ func (r *InstallationReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	log.V(Log4Debug).Info("Reconciliation complete: A porter agent has been dispatched to apply changes to the installation.")
-	if r.PorterGRPCClient != nil {
-		return r.CheckOrCreateInstallationOutputsCR(ctx, log, inst)
+	var conn *grpc.ClientConn
```
sgettys, line 176: This is a dupe of 127:135, can we just have CheckOrCreateInstallationOutputsCR handle the connection establishment as well?

controllers/installation_controller.go
sgettys: I agree, this is not great. Can you at least put a TODO in here to come back and clean this up? I can cut an issue for it too so we don't lose it. Honestly I feel like this is moving more and more towards having the controller actually manage the grpc server itself then it will know the context, it will be able to properly handle config for the server, and optionally start it based on a flag.

controllers/generate.go
```diff
@@ -1,3 +1,4 @@
 package controllers
 
 //go:generate mockery --name=PorterClient --filename=grpc_mocks.go --outpkg=mocks --output=../mocks/grpc
+//go:generate mockery --name=ClientConn --filename=clientconn_mocks.go --outpkg=mocks --output=../mocks/grpc
```
troy0820, line 4: Needed for tests

controllers/types.go
```diff
@@ -11,3 +11,7 @@ type PorterClient interface {
 	ListInstallations(ctx context.Context, in *installationv1.ListInstallationsRequest, opts ...grpc.CallOption) (*installationv1.ListInstallationsResponse, error)
 	ListInstallationLatestOutputs(ctx context.Context, in *installationv1.ListInstallationLatestOutputRequest, opts ...grpc.CallOption) (*installationv1.ListInstallationLatestOutputResponse, error)
 }
+
+type ClientConn interface {
+	Close() error
+}
```
troy0820, line 17: Necessary to abstract to allow for unit testing. The real grpc Client will need this to close the connection.

main.go
```diff
@@ -66,22 +63,12 @@ func main() {
...
-	if conn != nil {
-		client = porterv1alpha1.NewPorterClient(conn)
-	}
 	if err = (&controllers.InstallationReconciler{
 		Client:           mgr.GetClient(),
-		PorterGRPCClient: client,
 		Recorder:         mgr.GetEventRecorderFor("installation"),
 		Log:              ctrl.Log.WithName("controllers").WithName("Installation"),
 		Scheme:           mgr.GetScheme(),
+		CreateGRPCClient: controllers.CreatePorterGRPCClient,
```
troy0820, line 71: This sets this to the controller but can be overridden in tests.  Scoped to the controller to allow for parallel testing if necessary.

controllers/installation_controller.go
```diff
@@ -32,9 +35,9 @@ const (
 type InstallationReconciler struct {
 	client.Client
 	Log              logr.Logger
-	PorterGRPCClient PorterClient
```
troy0820, line 35: Client is now not scoped to the reconciler but has a method that will create one during the invocation of the parent method.  This method makes this a little more dynamic with how we wish to test this now that we moved the instantiation to the method `CheckOrCreateInstallationOutputs`

This interface is still available, but it doesn't hang off the reconciler and is injected with patching the function we now use that is scoped to the controller.

controllers/installation_controller.go
```diff
@@ -32,9 +35,9 @@ const (
...
 	Recorder         record.EventRecorder
 	Scheme           *runtime.Scheme
+	CreateGRPCClient func(ctx context.Context) (porterv1alpha1.PorterClient, ClientConn, error)
```
troy0820, line 40: Needed as we moved this to the parent function when doing outputs in the reconciler. This method can be overridden when testing and the interfaces are all mocked to show code path.

controllers/installation_controller.go
```diff
@@ -165,21 +166,26 @@ func (r *InstallationReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-		return r.CheckOrCreateInstallationOutputsCR(ctx, log, inst)
-	}
-	return ctrl.Result{}, nil
+	log.V(Log4Debug).Info(fmt.Sprintf("performing installation outputs for %s", inst.Name))
+	return r.CheckOrCreateInstallationOutputsCR(ctx, log, inst)
 }
 
 func (r *InstallationReconciler) CheckOrCreateInstallationOutputsCR(ctx context.Context, log logr.Logger, inst *v1.Installation) (ctrl.Result, error) {
 	// NOTE: May not want to requeue if this fails
+	porterGRPCClient, conn, err := r.CreateGRPCClient(ctx)
```
sgettys, line 175: nit: Guard check in case r.CreateGRPCClient is undefined. Shouldn't hit that though since this is part of the instantiation

sgettys (approved):

controllers/installation_controller.go
troy0820: Guarded against in line 175. Resolving.

controllers/installation_controller.go
troy0820: Added TODO, resolving.

PR #331, opened 2024-06-20
Delete without uninstall

### LINKED ISSUE #297: Deleting installations should uninstall and keep the Kubernetes behavior consistent with resource garbage collection

troy0820:
# What does this change
This allows deletion with uninstall and deletion with "orphan" by way of deletion policy.  

- [x] Add method for shouldOrphan 
- [x] Add check for shouldUninstall to check the deletion policy

# What issue does it fix
#297 again 

# Notes for the reviewer

# Checklist
- [x] Did you write tests?

### DIFF
--- a/controllers/installation_controller.go
+++ b/controllers/installation_controller.go
@@ -133,6 +133,10 @@ func (r *InstallationReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		err = r.uninstallInstallation(ctx, log, inst)
 		log.V(Log4Debug).Info("Reconciliation complete: A porter agent has been dispatched to uninstall the installation.")
 		return ctrl.Result{}, err
+	} else if r.shouldOrphan(inst) {
+		log.V(Log4Debug).Info("Reconciliation complete: Your installation is being deleted. Please clean up installation resources")
+		err = removeFinalizer(ctx, log, r.Client, inst)
+		return ctrl.Result{}, err
 	} else if isDeleted(inst) {
 		// This is installation without a finalizer that was deleted We remove the
 		// finalizer after we successfully uninstall (or someone is manually cleaning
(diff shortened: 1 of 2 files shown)


controllers/installation_controller.go
```diff
@@ -401,7 +405,11 @@ func (r *InstallationReconciler) saveStatus(ctx context.Context, log logr.Logger
 
 func (r *InstallationReconciler) shouldUninstall(inst *v1.Installation) bool {
 	// ignore a deleted CRD with no finalizers
-	return isDeleted(inst) && isFinalizerSet(inst)
+	return isDeleted(inst) && isFinalizerSet(inst) && inst.GetAnnotations()[v1.PorterDeletePolicyAnnotation] == v1.PorterDeletePolicyDelete
+}
+
+func (r *InstallationReconciler) shouldOrphan(inst *v1.Installation) bool {
+	return isDeleted(inst) && isFinalizerSet(inst) && inst.GetAnnotations()[v1.PorterDeletePolicyAnnotation] == v1.PorterDeletePolicyOrphan
```
troy0820, line 412: I can test this.

sgettys (approved):

PR #332, opened 2024-06-20
WIP: Create grpc deployment when starting operator 

### LINKED ISSUE #308: Bundle grpc server with controller-manager in the operator pod spec
**Is your feature request related to a problem? Please describe.**
We have the capability to ship the grpc server along with the operator in one yaml instead of trying to drive them separately.  Currently we do not ship the grpc server and it doesn't allow the installation outputs to work without them.

**Describe the solution you'd like**
Create the yaml that is going to be deployed and allow that to be used with the deployment of the operator.  That deployment should include the grpc service to be used by the porter operator within the context of creating a porter operator.  The binary for this we can create a generic grpc server and inject the service necessary as well as provide the service during deployment into the porter-operator namespace.  

**Describe alternatives you've considered**
What is described above but this effort will have to come in phases as this will require that porter main can install this properly

**Additional context**
None

troy0820:
# What does this change
Creates the GRPC deployment, configmap, service when starting the operator 

- [x] Create go routines which will handle the deployment creation asynchronously
- [ ] 

# What issue does it fix
Closes #308

# Notes for the reviewer

# Checklist
- [ ] Did you write tests? _Not yet_


main.go
```diff
@@ -116,6 +126,61 @@ func main() {
...
+		})
+
+		go func() {
+			if err := g.Wait(); err != nil {
+				setupLog.Error(err, "error with async operation of creating grpc deployment")
+				os.Exit(1)
+			}
+			setupLog.Info("grpc server has been created")
+		}()
+	}
```
troy0820, line 182: This is dependent on the resources like configmap being there first, etc.  So the ordering of these async methods may need to be:
1. Check/Create configmap
2. Check/Create service
3. Check/Create deployment

So this will probably being one method.  

Also, this required more from an RBAC perspective.

troy0820: When the operator starts, the cache is not loaded to GET any resource.  The create will have to happen and if it exists (conflict), we will assume this will be there.  

This doesn't ensure that someone will not delete the deployment/configmap/service.

troy0820: Resolving.

controllers/types.go
```diff
@@ -15,3 +26,118 @@ type PorterClient interface {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{
+					"app": "porter-grpc-service",
+				},
+			},
+			Spec: corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name:  "porter-grpc-service",
+						Image: "ghcr.io/getporter/server:v1.1.0",
```
sgettys, line 55: Can me make the image version a const? I think it's fine to have the server version compiled into the operator controller to ensure compatibility but it should be easy to upgrade when we want to release a new version

controllers/types.go
```diff
@@ -15,3 +26,118 @@ type PorterClient interface {
...
+				TargetPort: intstr.FromInt(3001),
+				Port:       int32(3001),
+			},
+		},
+		Selector: map[string]string{"app": "porter-grpc-service"},
+		Type:     corev1.ServiceTypeClusterIP,
+	},
+}
+
+var GrpcConfigMap = &corev1.ConfigMap{
```
sgettys, line 125: This is where I think the complexity is for managing the grpc server as part of the controller. The porter config that this uses HAS to be the porter config that the installations used when running so that means it needs to use the porter config for the namespace. We put the restriction in the only allow a single porter config per namespace instead of a porter config to be defined for every installation. The operator sets up a default config if one doesn't exist. This service should be using the porter config that's applied to the namespace, either the default one or the user provided one. It also has to track if that PorterConfig ever changes and then reload with the new config so that it can hit the backing stores where the installations actually live.

Btw @schristoff we should add support for moving porter storage data from one backend to another. This will "just work" for secrets as long as the secrets are managed externally for the new store like they should be but for storage backend AFAIK there's no way to move from one DB to another if that changes in the porter config and we should definitely support that!

controllers/types.go
troy0820: I can update this.

controllers/types.go
troy0820: > The porter config that this uses HAS to be the porter config that the installations used when running so that means it needs to use the porter config for the namespace. We put the restriction in the only allow a single porter config per namespace instead of a porter config to be defined for every installation. The operator sets up a default config if one doesn't exist. This service should be using the porter config that's applied to the namespace, either the default one or the user provided one. It also has to track if that PorterConfig ever changes 

That means we can't do this implementation because this is trying to install the grpc server before we even do an installation.  If the installation needs to happen to success to then create a deployment/configmap/service to get the Porter config in the namespace, (that resource isn't created until something makes it get created) then we will have to do that after the first installation.  Relying on the default Porter config at runtime seems a little tricky as that default is making assumption around the installation resource process.  

What I can do is move this to be done once at setup during the first installation and dynamically create the configmap that will map to the porter config by the installation we have in the namespace once the installation is complete.

sgettys: This is really interesting because of how the Porter config gets resolved for an AgentAction:
https://github.com/getporter/operator/blob/0783aa453566706973fe5909b8eec20c6b245559/controllers/agentaction_controller.go#L554
We probably need to create a "default" grpc server in the operator namespace then have namespace specific servers IF a PorterConfig is specified for that namespace BUT not for the system..... This just feels gross.... But the resolvePorterConfig should be able to handle checking if a grpc server exists that will be able to handle the AgentAction based on the PorterConfig that is selected to run that AgentAction

controllers/types.go
```diff
@@ -5,6 +5,17 @@ import (
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
+)
+
+const (
+	PorterNamespace = "porter-operator-system"
```
troy0820, line 17: This is already defined in controllers/installation_controller.go

troy0820:
After discussion, we agreed that this should be created on a per PorterConfig basis.  If none exists, fall back to the default one that is created in the porter-operator-system namespace.

